PR #8, opened 2016-03-10
write to sys.stdout in python 2.X, not sys.stdout.buffer

jsfenfen:
Writing to sys.stdout.buffer dies on 2.7.11; use sys.stdout


jsvine:
Big thanks for catching these bugs! :tada:

jsvine:
Fixes now in v0.4.1 and available on PyPi.

Issue #21, opened 2017-02-22
page height is a float in python 2.7? 


jsfenfen:
I'm seeing this with a current version of pdf plumber. Seems to only apply to some pages? 

...pdfplumber_env/lib/python2.7/site-packages/pdfplumber/page.py", line 65, in process_object
    attr["top"] =h - attr["y1"]
TypeError: unsupported operand type(s) for -: 'float' and 'Decimal'

jsvine:
Thanks for flagging! Do you have an example PDF that throws this error? For the purposes of reproducing the bug, add a test, and fix it.

jsfenfen:
Here's one: 
[150109DSP-Milw-505-90D.pdf](https://github.com/jsvine/pdfplumber/files/794587/150109DSP-Milw-505-90D.pdf)

jsvine:
Super, thanks!

jsvine:
Many thanks again. Fixed, tested, and incorporated in v0.5.0/#23

PR #22, opened 2017-02-22
decimalize page height-it seems to sometimes come as a float

jsfenfen:
See: https://github.com/jsvine/pdfplumber/issues/21 . Not sure this is the best fix, but seems to work for me.


jsvine:
Thanks! Fixed in https://github.com/jsvine/pdfplumber/pull/23

PR #48, opened 2018-01-20
Update setup.py so it doesn't fail on pip install pycrypto

danshorstein:
replaced pycrypto with pycryptodome dependency so install doesn't fail on pip install


jsvine:
Good call, and thanks!

jsvine:
Now available in `v0.5.7`.

PR #77, opened 2018-08-13
resolve page rotation when PDFObjRef is returned

meldonization:
pdfplumber ERROR message without the fix.
```
>>> pdf = pdfplumber.open(fpath)
>>> page_num = len(pdf.pages)
Traceback (most recent call last):
  File "<input>", line 1, in <module>
  File "~/.pyenv/versions/3.5.2/lib/python3.5/site-packages/pdfplumber/pdf.py", line 58, in pages
    p = Page(self, page, page_number=page_number, initial_doctop=doctop)
  File "~/.pyenv/versions/3.5.2/lib/python3.5/site-packages/pdfplumber/page.py", line 19, in __init__
    self.rotation = self.page_obj.attrs.get("Rotate", 0) % 360
TypeError: unsupported operand type(s) for %: 'PDFObjRef' and 'int'
```

### DIFF
--- a/pdfplumber/page.py
+++ b/pdfplumber/page.py
@@ -16,7 +16,10 @@ class Page(Container):
         self.pdf = pdf
         self.page_obj = page_obj
         self.page_number = page_number
-        self.rotation = self.page_obj.attrs.get("Rotate", 0) % 360
+        _rotation = self.page_obj.attrs.get("Rotate", 0)
+        if not isinstance(_rotation, int):
+            _rotation = _rotation.resolve()
+        self.rotation = _rotation % 360
         self.page_obj.rotate = self.rotation
         self.initial_doctop = self.decimalize(initial_doctop)
 


meldonization:
After posting this issue, I've found many more errors with regard to returned PDFObjRef in function `utils.decimalize`.
So I would suggest add a function in utils named `resolve_pdfobjref` which resolves the true value of any pdf reference object.
Many thanks :p

jsvine:
Thanks for flagging. Could you provide an example of a PDF that causes this problem? And others that are causing "many more errors with regard to returned PDFObjRef"?

meldonization:
@jsvine  Attached is the pdf file. 
[pdf_with_pdfobjrefs.pdf](https://github.com/jsvine/pdfplumber/files/2289858/5983f7fecb379e697f625c90.pdf)

PR #100, opened 2019-02-07
Option to return the matrix and bounding boxes of objects

CorSteging:
The translation matrix and bbox of objects on a page are ignored when parsed. This change adds an option to return the matrices and bounding boxes of the objects. By default they are still ignored.

### DIFF
--- a/pdfplumber/page.py
+++ b/pdfplumber/page.py
@@ -27,6 +27,8 @@ class Page(Container):
         self.cropbox = self.decimalize(resolve_all(cropbox)) if cropbox is not None else None
         self.mediabox = self.decimalize(resolve_all(mediabox) or self.cropbox)
         m = self.mediabox
+        self.return_matrix = False
+        self.return_bbox = False
 
         if self.rotation in [ 90, 270 ]:
             self.bbox = self.decimalize((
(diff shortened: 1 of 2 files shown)


jsvine:
Thanks, @CorSteging, and my apologies for the slowness in responding. Could you explain the use-cases for these options? The information in `bbox`, for example, would be redundant with the `x0, y0, x1, y1` information, no?

CorSteging:
Thank you for the response @jsvine. The most important part of this PR is the inclusion of the translation matrix. Without this parameter, it can be impossible to determine the exact location of the words/characters on PDFs with an inherent translation matrix. 

The information provided by the `bbox` attribute can also be retrieved through the `x0, y0, x1, y1` variables, so in that sense it is redundant. In a lot of cases, however, the complete bbox of a word is required, making quite convenient to have the option to return the bboxes directly. Especially considering the fact that these bounding boxes are already there, but simply ignored.

jsvine:
Thank you for clarifying that. I appreciate it! I've been a bit busy, but I'll take this into consideration.

jsvine:
Just pushed an attempt at making the current transformation matrix (CTM) accessible for characters, and introduced a class (`pdfplumber.ctm.CTM`) for calculating its scale, skew, and translation: https://github.com/jsvine/pdfplumber/commit/ae6f99e691de72987ab8166403a39776a41d4c30

For now, the "matrix" property of a `char` will be accessible by default, rather than introducing a `PDF(...)` initialization parameter to control whether or not it appears. But finer-grained controls over what properties are extracted may be a nice feature in the future.

To test this out, you can install the unreleased code via `pip install -e git+https://github.com/jsvine/pdfplumber.git@feature-ctm`

jsvine:
Just landed in v0.7.0!

PR #107, opened 2019-03-30
sort by y coordinate

ldhsight:
sort tables by bbox[1] (y coordinate), this will make us extracting tables orderly.
As before, we sometimes got tables by func extract_tables with wrong order.For example,
if a table separated into 2 or more parts in different pages, the tables we extract will be sorted
by length desc, in this way we can not extract them in their raw order .So i think we should extract
then by y coordinate ,that is the right way to avoid confused order after extracting

### DIFF
--- a/pdfplumber/page.py
+++ b/pdfplumber/page.py
@@ -176,7 +176,8 @@ class Page(Container):
 
     def extract_tables(self, table_settings={}):
         tables = self.find_tables(table_settings)
-        return [ table.extract() for table in tables ]
+        tables = sorted(tables, key=lambda x: x.bbox[1])
+        return [ table.extract() for table in tables]
 
     def extract_table(self, table_settings={}):
         tables = self.find_tables(table_settings)


jsvine:
Hi, @ldhsight. I think there may be a slight misunderstanding. 

The sort-by-length is a heuristic used only by `.extract_table(...)` (singular), on the thought that if a user is requesting just one table, they're probably looking for the largest one.

On the other hand, `.extract_tables(...)` (plural) returns the tables sorted by `x0, y0`, per this line: https://github.com/jsvine/pdfplumber/blob/3262de7486559fd2f4c710be1143cb126cc8394b/pdfplumber/table.py#L312

Changing the sort order from `x`-first to `y`-first might cause some breaking changes for users, but I might think about it for `v0.6.0`. In the meantime, you can always custom-sort the result of `.extract_tables(...)` to your preference.

Or am I misunderstanding this PR?

ldhsight:
ok, thanks for your response, it's reasonable for the opinion 'looking for the largest one' and i think so as well, haha, thanks again (^_^)

jsvine:
Sounds good!

PR #108, opened 2019-04-04
added optional support for cdecimal to improve performance

AdamRi:
`cdecimal` is much faster than `decimal` in Python 2 (it was later integrated into Python 3).
This pull request lets pdfplumber import `cdecimal` if it exists.

### DIFF
--- a/pdfplumber/cli.py
+++ b/pdfplumber/cli.py
@@ -2,7 +2,10 @@
 import pdfplumber
 import argparse
 from itertools import chain
-from decimal import Decimal, ROUND_HALF_UP
+try:
+    from cdecimal import Decimal, ROUND_HALF_UP
+except ImportError:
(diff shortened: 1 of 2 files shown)


jsvine:
Thank you for raising this suggestion, @AdamRi. Personally, I'm on the fence. Clearly, it's a performance improvement for Python 2. On the other hand, Python 2 is on its way out, and I'm wary of adding code specific to it unless necessary. About how much speedup are you seeing with this change? And how crucial is it to your work?

AdamRi:
Our code is running more than twice as fast with this. Python 2 will cease to have support by 2020, and we are working on migrating to Python 3, but in the meantime such an improvement is indeed crucial for us.

jsvine:
Roger that. Merged and now available in `v0.5.12`.

Issue #128, opened 2019-07-19
Adding a delimiter option for extract_text()


shivampatel8:
Currently it functions as adding a space between the words that exceed the tolerance. Looking at the code It would be quite simple to add the delimiter option. I would like to do this.

na4raitda:
I really need this option...
For I python newbie, it's difficult to parse since there are whitespaces everywhere.

jsvine:
Thanks, @shivampatel8 for raising this, and my apologies for the delay in responding. I'm trying to catch up on old issues. Are you still interested to contribute this code? If so, feel free to submit a PR. Otherwise, I will add it to my todo list.

shivampatel8:
I have a pull request open for this waiting for your approval.

jsvine:
Ah, right! My apologies, I didn’t see it referenced here. Closing this issue for now; we can continue the discussion in the PR. Thanks!

PR #129, opened 2019-07-19
Action on Issue #128

shivampatel8:
Added delimiter option to utils.py and page.py with default option as DEFAULT_DELIMITER which is a single space


jsvine:
Hi @shivampatel8! I'm not opposed to this PR, but I'd like to understand more about it. Do you have an example of when a custom delimiter would be useful?

Also, if the `" "` delimiter is going to be overridden, should there also be an option to override the `\n` joiner here?: https://github.com/jsvine/pdfplumber/pull/129/files#diff-fa17a09eee3f5f7b067fc8eeb37ba2adR182

Finally, I think the `DEFAULT_DELIMITER` constant name is a bit vague, and possibly confusing, since I think many people will thing it describes a *table* delimiter. Since this setting is basically limited to this one function, perhaps it's best to ditch the constant, and slightly rename the parameter. Perhaps `word_joiner = " "` ... and then, per above, could also add `line_joiner = "\n"`?

shivampatel8:
So I was working on parsing some documents. The initial hope was that I would be able to use the table parser and that would give me everything but that did not work out so I made a work around and It was based off of some regex matching of dates. This gave me all of the lines in the table but there was no way to identify the columns since everything was separated by spaces. So I went into the source code and messed around a little to see if I could get that to workout and that is where I came up with the idea.

The exact use would be for when the x_tolerance is high to separate big white spaces(also if you have multi-word statements which are getting clumped together since the word separator is just a space.

In regards to the line_joiner. I did think of this as well afterwards but was waiting for feedback. I will get these changes up ASAP.

shivampatel8:
Files updated based on your requests

jsvine:
Thank you, @shivampatel8! There are a few issues with this PR (mainly, that it fails the test suite due to a couple of `SyntaxError: non-default argument follows default argument` exceptions), but this is a great start, and I can work on incorporating what you’ve done here. One thing that I think we’ll still want to decide on: How should these options be reflected in `extract_words`? Should it also be possible to pass a custom word-delimiter to that function? I think so, and can work on incorporating that.

Issue #168, opened 2020-01-08
Customizing underlying pdfminer laparams fails


frascuchon:
I'm trying to customize the underlying pdfminer text extract process for better tune my text extraction pipeline by setting some `LAParams` values falling into an error:

```shell
Traceback (most recent call last):
  File "/usr/local/anaconda3/envs/biome/lib/python3.7/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/usr/local/anaconda3/envs/biome/lib/python3.7/unittest/case.py", line 615, in run
    testMethod()
  File "/Users/frascuchon/recognai/pdfplumber/tests/test_laparams_customization.py", line 20, in test_load_with_custom_laparams
    print(first_page.chars)
  File "/Users/frascuchon/recognai/pdfplumber/pdfplumber/container.py", line 35, in chars
    return self.objects.get("char", [])
  File "/Users/frascuchon/recognai/pdfplumber/pdfplumber/page.py", line 66, in objects
    self._objects = self.parse_objects()
  File "/Users/frascuchon/recognai/pdfplumber/pdfplumber/page.py", line 167, in parse_objects
    process_object(obj)
  File "/Users/frascuchon/recognai/pdfplumber/pdfplumber/page.py", line 140, in process_object
    for k, v in obj.__dict__.items()
  File "/Users/frascuchon/recognai/pdfplumber/pdfplumber/page.py", line 141, in <genexpr>
    if k not in IGNORE)
KeyError: 'index'
```

You can check on your own with this code snippet:
```python
import pdfplumber

with pdfplumber.open(path, laparams=dict(line_margin=0.2)) as pdf:
    print(f"Found {len(pdf.pages)} pages")
    first_page = pdf.pages[0]
    print(first_page.chars)
```

jsvine:
Fixed in https://github.com/jsvine/pdfplumber/commit/df00787adae64af94f899b83e5909fefd53a5574 and now available in v0.5.16.

PR #169, opened 2020-01-08
feat: allow parameterize laparams

frascuchon:
This PR resolves issue #168

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 /notebooks
 .ipynb_checkpoints
 .DS_Store
+.idea
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
@@ -59,3 +60,5 @@ docs/_build/
 
 # PyBuilder
 target/
+
+
--- a/pdfplumber/page.py
+++ b/pdfplumber/page.py
@@ -6,10 +6,12 @@ from copy import copy
 from pdfminer.pdftypes import resolve_all
 from six import string_types
 import re
+
 lt_pat = re.compile(r"^LT")
 
+
 class Page(Container):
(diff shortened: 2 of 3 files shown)


jsvine:
Hi @frascuchon, and many thanks for your proposal! I think it’s a good idea to allow for laparams customization. I don’t, however, see where in this PR that’s implemented. Am I missing something?

Part of my difficulty in reading this PR is that there seem to be many code-formatting changes introduced at the same time. While I don’t disagree with most of them, they add a lot of noise to the PR, making it difficult to identify the parts that have meaningfully changed. Can you resubmit the PR without stylistic/formatting changes? (I’d prefer to handle the formatting topic in a separate PR; the approach makes sense to me, but I think it should be consistent throughout the whole project, rather than confined to one file.)

p.s.,. Smart to use the conversion keys as the attributes filter! Great catch.

jsvine:
Ah, I now see what I was missing: It is already possible to pass custom LAParams to pdfplumber; but when you do, it throws an error (as you note in #168). Your solution is to use the `CONVERSIONS` dict instead of the `IGNORE` list as the initial filter in `Page.process_object(obj)`. That does seem to address the error — thanks! — but I want to take a slightly closer look at what was causing the error in the first place. I’ll update this thread when I do.

jsvine:
Reviewing the cause of that error, it seems relatively benign (the addition of an `index` property to char objects), and your fix seems like a good one. I have incorporated it in https://github.com/jsvine/pdfplumber/commit/df00787adae64af94f899b83e5909fefd53a5574 and credited you. (Thanks!) The fix is now available in v0.5.16.

Closing this PR, but if you want to discuss code-formatting, feel free to open an issue on the topic.

PR #179, opened 2020-02-05
Fix #124 opening PDF with bytes stream

### LINKED ISSUE #124: '_io.BytesIO' object has no attribute 'name'
The `to_image()` method does not seem to work if the `pdfplumber.PDF` object was created using a `BytesIO` stream. The rest of the functionality seems unaffected.

The problem seems to arise in the call to `wand.image.Image()` in the `get_page_image()` function in `display.py`. This image function have the ability to take file objects using the `file` argument explained [here](http://docs.wand-py.org/en/0.4.1/wand/image.html) but `get_page_image()` only ever uses the `filename` parameter. Line 42 of the `PageImage` class is also looking for the name of the stream, but BytesIO objects do not have a name. Extracting characters, rectangles etc. can still be done with these BytesIO objects.

The MWE:

```python3
import pdfplumber
from io import BytesIO

file_path = "file.pdf"

# This example succesfully extracts chars and makes an image
file_like_object = open(file_path, "rb") # _io.BufferedReader object
first_page = pdfplumber.load(file_like_object).pages[0]
chars = first_page.chars
im = first_page.to_image()

# This example succesfully extracts chars but does not make an image
file_like_object.seek(0)
different_file_like_object = BytesIO(file_like_object.read()) # _io.BytesIO object
first_page_2 = pdfplumber.load(different_file_like_object).pages[0]
chars_2 = first_page_2.chars
im = first_page_2.to_image()
```

Gives the error:
```pytb
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-9-3f3cda6c1277> in <module>
      5 first_page_2 = pdfplumber.load(different_file_like_object).pages[0]
      6 chars_2 = first_page_2.chars
----> 7 im = first_page_2.to_image()

~/web_app/lib/python3.6/site-packages/pdfplumber/page.py in to_image(self, **conversion_kwargs)
    256         if "resolution" not in conversion_kwargs:
    257             kwargs["resolution"] = DEFAULT_RESOLUTION
--> 258         return PageImage(self, **kwargs)
    259 
    260 class DerivedPage(Page):

~/web_app/lib/python3.6/site-packages/pdfplumber/display.py in __init__(self, page, original, resolution)
     40         if original == None:
     41             self.original = get_page_image(
---> 42                 page.pdf.stream.name,
     43                 page.page_number - 1,
     44                 resolution

AttributeError: '_io.BytesIO' object has no attribute 'name'
```

Not sure how best to fix this issue.

cheungpat:
My app fetches PDF from HTTP and it creates a PDF object using a `io.BytesIO` stream. Since the stream is not opened from a file, it does not have the `name` property which `PageImage` expects it to have.

This commit fixes the issue #124 by using existing stream object to create a wand Image. The stream is already read by `PDFParser` so we need to seek backwards to the beginning.

This should work for both file-stream and byte-stream.

### DIFF
--- a/pdfplumber/display.py
+++ b/pdfplumber/display.py
@@ -17,12 +17,13 @@ DEFAULT_STROKE = COLORS.RED + (200,)
 DEFAULT_STROKE_WIDTH = 1
 DEFAULT_RESOLUTION = 72
 
-def get_page_image(pdf_path, page_no, resolution):
+def get_page_image(stream, page_no, resolution):
     """
     For kwargs, see http://docs.wand-py.org/en/latest/wand/image.html#wand.image.Image
     """
-    page_path = "{0}[{1}]".format(pdf_path, page_no)
-    with wand.image.Image(filename=page_path, resolution=resolution) as img:
+    stream.seek(0)
+    with wand.image.Image(file=stream, resolution=resolution) as pages:
+        img = wand.image.Image(image=pages.sequence[page_no])
         if img.alpha_channel:
             img.background_color = wand.image.Color('white')
             img.alpha_channel = 'background'
@@ -39,7 +40,7 @@ class PageImage(object):
         self.page = page
         if original == None:
             self.original = get_page_image(
-                page.pdf.stream.name,
+                page.pdf.stream,
                 page.page_number - 1,
                 resolution
             )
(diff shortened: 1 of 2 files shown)


jsvine:
Thank you @cheungpat! On a first glance, this seems like a great fix, and thank you for adding a test. I hope to take a closer look soon and, likely, merge.

jsvine:
Thanks again for this, @cheungpat! Merging.

PR #200, opened 2020-04-07
Add missing object type `rect_edge` to `obj_to_edges()`

samkit-jain:
Method `obj_to_edges()` was added in https://github.com/jsvine/pdfplumber/commit/d68e8813be6ab2ff74d8a8048f3313d48ba91788 but missed out on adding `rect_edge` as a valid key. `rect_edge` is being used in https://github.com/jsvine/pdfplumber/blob/dead89c86d1cac5535d40326adb8ccdcde7d1b5b/pdfplumber/utils.py#L423-L452 and https://github.com/jsvine/pdfplumber/blob/dead89c86d1cac5535d40326adb8ccdcde7d1b5b/pdfplumber/cli.py#L42

Also updated an existing test `test_pr_138()` with a call to `extract_tables` that would raise an error `KeyError: 'rect_edge'` when run on the `master` branch.

### DIFF
--- a/pdfplumber/utils.py
+++ b/pdfplumber/utils.py
@@ -460,6 +460,7 @@ def obj_to_edges(obj):
     return {
         "line": lambda x: [ line_to_edge(x) ],
         "rect": rect_to_edges,
+        "rect_edge": rect_to_edges,
         "curve": curve_to_edges,
     }[obj["object_type"]](obj)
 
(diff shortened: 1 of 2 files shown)


jsvine:
A big thanks, @samkit-jain! That's a great catch, and I really appreciate you updating that test. Merging.

Issue #203, opened 2020-04-22
metadata parse bug & bbox decimalize bug


meldonization:
> pdf = pdfplumber.open('metadata_bug.pdf')
```
Traceback (most recent call last):
  File "<input>", line 11, in <module>
  File ".venv/lib/python3.6/site-packages/pdfplumber/utils.py", line 82, in decode_text
    return ''.join(PDFDocEncoding[o] for o in ords)
  File ".venv/lib/python3.6/site-packages/pdfplumber/utils.py", line 82, in <genexpr>
    return ''.join(PDFDocEncoding[o] for o in ords)
  File ".venv/lib/python3.6/site-packages/pdfplumber/utils.py", line 81, in <genexpr>
    ords = (ord(c) if type(c) == str else c for c in s)
TypeError: ord() expected a character, but string of length 4 found
```

pdf: [metadata_bug.pdf](https://github.com/jsvine/pdfplumber/files/4514320/metadata_bug.pdf)

> pdf = pdfplumber.open('metadata_bug.pdf')
> pdf.pages[0].objects
```
Traceback (most recent call last):
  File "<input>", line 1, in <module>
  File ".venv/lib/python3.6/site-packages/pdfplumber/page.py", line 66, in objects
    self._objects = self.parse_objects()
  File ".venv/lib/python3.6/site-packages/pdfplumber/page.py", line 155, in parse_objects
    process_object(obj)
  File ".venv/lib/python3.6/site-packages/pdfplumber/page.py", line 152, in process_object
    process_object(child)
  File ".venv/lib/python3.6/site-packages/pdfplumber/page.py", line 128, in process_object
    for k, v in obj.__dict__.items()
  File ".venv/lib/python3.6/site-packages/pdfplumber/page.py", line 129, in <genexpr>
    if k in CONVERSIONS_KEYS)
  File ".venv/lib/python3.6/site-packages/pdfplumber/page.py", line 47, in decimalize
    return utils.decimalize(x, self.pdf.precision)
  File ".venv/lib/python3.6/site-packages/pdfplumber/utils.py", line 115, in decimalize
    return type(v)(decimalize(x, q) for x in v)
  File ".venv/lib/python3.6/site-packages/pdfplumber/utils.py", line 115, in <genexpr>
    return type(v)(decimalize(x, q) for x in v)
  File ".venv/lib/python3.6/site-packages/pdfplumber/utils.py", line 117, in decimalize
    return _decimalize(v, q)
[...]

PR #204, opened 2020-04-22
Fix metadata bug and decimalize bug on issue #203

meldonization:
see also in issues #203


jsvine:
Thanks, @meldonization! I think we probably want to resolve the PDF object *before* it even gets to `._decimalize`. But the bug you've reported in #203 is helpful, and enough for me to get started looking for an earlier fix.

jsvine:
Thanks again, @meldonization! Closing for now, per https://github.com/jsvine/pdfplumber/issues/203.

PR #213, opened 2020-05-20
Add stroking and non-stroking colors for objects

### LINKED ISSUE #148: Is it possible to extract Font Color?
Hi, I was just wondering if it's possible to extract the font color of characters from the PDF as they are normally available in the font-specs properties of the files.

I am aware of the `fontname` attribute of `pdf.chars` but is there any setting to extract other font properties as well?

@jsvine is this available in pdfplumber?

idan-david:
Fixes #148

### DIFF
--- a/pdfplumber/page.py
+++ b/pdfplumber/page.py
@@ -132,6 +132,11 @@ class Page(Container):
             attr["object_type"] = kind
             attr["page_number"] = pno
 
+            if hasattr(obj, "graphicstate"):
+                gs = obj.graphicstate
+                attr['stroking_color'] = gs.scolor
+                attr['non_stroking_color'] = gs.ncolor
+
             if hasattr(obj, "get_text"):
                 attr["text"] = obj.get_text()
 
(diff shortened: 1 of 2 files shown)


jsvine:
Thanks for sharing this, @idan-david! `pdfplumber` already picks up on some color information — see, e.g., here: 

https://github.com/jsvine/pdfplumber/blob/df00787adae64af94f899b83e5909fefd53a5574/tests/test-basics.py#L59-L63

... so I'm interested to learn more about what this PR fixes/adds.

Would you be able to add a test that fails if run with the current `master` branch but passes with this PR's changes?

idan-david:
I have added a test.
Unlike `LTRect`, `LTChar` objects in pdfminer.six do not have a "non_stroking_color" property, so no color information is extracted for them. The color information lies in the `graphic_state` property.

jsvine:
Many thanks! Added, merged, and now available in `v0.5.22`.

Issue #28, opened 2017-03-08
word-level font names and heights


jsfenfen:
Having a font for an entire word helps parsing. A lot. Height also helps some. 

I took a crack at this here, with some settings. Defaults also may need adjustment.

If you've got thoughts, @jsvine, lemme know and I can clean this up into a pr. Haven't gotten the testing set up yet. 

https://github.com/jsfenfen/pdfplumber/commit/847a3bbe637e4a3c6bf7a8151eeccf166b3d4fdf

jsfenfen:
I guess with word heights I'm going back and fourth on averaging them or taking the mode; left the latter in for the moment.

jsvine:
Thanks! I like this. For testing's sake: Do you have shareable examples of PDFs where chars that *should* belong to the same word either have different heights or fontnames?

jsfenfen:
So I still haven't heard back about the files that originally required this. I could pretty easily just *make up* a sample pdf that failed the font height test, though obviously having an example would be better... The other time this stuff (can) come up is when the word tolerance is set too high and words run together inadvertently--though only if adjacent cells have different fonts. Will look around a bit.

jsvine:
No worries. Thinking through this a bit. I'm tempted to, by default, group words by fonts, size, *and* color. (Yes, upcoming versions of `pdfplumber` will include font color!) Boolean params could turn them off. I.e., defaults would be:

```python
def extract_words(chars,
  x_tolerance=DEFAULT_X_TOLERANCE,
  y_tolerance=DEFAULT_Y_TOLERANCE,
  keep_blank_chars=False,
  match_fontsize=True,
  match_fontcolor=True,
  match_fontname=True
)
```

That'd mean losing some of the flexibility of, e.g., `DEFAULT_FONT_HEIGHT_TOLERANCE`, but might make the options clearer. It'd also mean avoiding having to calculate the average/mode values for `tolerance`-ed attributes. For instance, this ...

```python
page.extract_words()
```

.... might return ...

```python
[ {
  "text": "Hello",
  "fontsize": 12,
  "fontname": "ArialBold",
  "fontcolor": "#000000"
} ]
```

... while ...

```python
page.extract_words(match_fontsize=False)
```

.... would return ...

```python
[ {
  "text": "Hello",
  "fontname": "ArialBold",
  "fontcolor": "#000000"
} ]
```

What do you think? Too inflexible?

jsfenfen:
I think that's great! 

[...]

PR #260, opened 2020-08-29
Refactor several complex methods and add `extra_attrs` to `.extract_words(...)`

jsvine:
This commit largely focuses on refactoring a few previously-complex methods. In the process, however, it also adds support for the long-requested ability to group word-characters on (and convey information about) extra attributes (such as `fontname` and `size`) (#28).

See CHANGELOG.md for more detailed list of changes.


CHANGELOG.md
```diff
@@ -2,6 +2,19 @@
 
 All notable changes to this project will be documented in this file. The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
+## [0.5.24] — Unreleased
+### Added
+- Added `extra_attrs=[...]` parameter to `.extract_text(...)` (c8b200) (#28)
```
samkit-jain: Can use hyperlinks for commit hash and issue number as it would make it easier to jump to those.

pdfplumber/utils.py
```diff
@@ -205,86 +205,103 @@ def bbox_to_rect(bbox):
...
 
-        return [process_word_chars(chars, upright) for chars in words]
+    x_tolerance = decimalize(x_tolerance)
+    y_tolerance = decimalize(y_tolerance)
 
-    chars_by_upright = {1: [], 0: []}
     words = []
-    for char in to_list(chars):
-        chars_by_upright[char.get("upright", 1)].append(char)
+    grouped = itertools.groupby(chars, itemgetter("upright", *extra_attrs))
```
samkit-jain: `itertools.groupby()` expects the iterable to be sorted. Should that matter here? I think so not.

README.md
```diff
@@ -100,7 +100,7 @@ The `pdfplumber.Page` class is at the core of `pdfplumber`. Most things you'll d
...
+|`.extract_words(x_tolerance=3, y_tolerance=3, keep_blank_chars=False, horizontal_ltr=True, vertical_ttb=True, extra_attrs=[])`| Returns a list of all word-looking things and their bounding boxes. Words are considered to be sequences of characters where (for "upright" characters) the difference between the `x1` of one character and the `x0` of the next is less than or equal to `x_tolerance` *and* where the `doctop` of one character and the `doctop` of the next is less than or equal to `y_tolerance`. A similar approach is taken for non-upright characters, but instead measuring the vertical, rather than horizontal, distances between them. The parameters `horizontal_ltr` and `vertical_ttb` indicate whether the words should be read from left-to-right (for horizontal words) / top-to-bottom (for vertical words). Changing `keep_blank_chars` to `True` will mean that blank characters are treated as part of a word, not as a space between words. Passing a list of `extra_attrs`  (e.g., `["fontname", "size"]` will restrict each words to characters that share exactly the same value for each of those attributes, and the resulting word dicts will indicate those attributes.|
```
samkit-jain: Can also mention that the valid attributes are the ones mentioned in https://github.com/jsvine/pdfplumber/blob/develop/README.md#char-properties

CHANGELOG.md
jsvine: Good call. I had hoped GitHub would do that automatically, but it appears not. Now added in 5d51940.

pdfplumber/utils.py
jsvine: This small-but-smart question led me more broadly to question assumptions about sort-order in `.extract_words`, and led me to (a) rewriting the logic again, in a hopefully more robust and clearly-defined way, and (b) adding a `use_text_flow` parameter. (Open to ideas for other names for that.) More details below, in 84f9181.

README.md
jsvine: Thanks, you're right that this should be more explicit about what an attribute can be. For concision's sake, though, I've decided just to hyperlink to that subsection, rather than adding more words.

pdfplumber/utils.py
jsvine: And, just to follow up on the original question: In order to make the order of the words returned by `.extract_words` predictable and consistent (top-down, and then either left-right or right-left, depending on `horizontal_ltr`), I don't think we want to sort by `extra_attrs`... but I'm open to other perspectives on this.

CHANGELOG.md
samkit-jain: Missed adding the `/commit` part in the URL which resulted in a 404 when opening the links. Fixed in https://github.com/jsvine/pdfplumber/pull/260/commits/cb924346bb5c144251dfa683bc2863f6e2237a34

jsvine (commented):
Ooof, I really should have caught those hyperlink problems. Thank you!

samkit-jain (approved):

Issue #277, opened 2020-09-27
Option to extract text of cells with leading and trailing white spaces preserved


AysegulKarcili:
Hi! First of all, thanks a lot for this library since it is excellent for extracting tables, being very easy and intuitive to use as well.

I am working on tables which are too long and split to multiple pages. Splits sometimes occurs between two table rows, but sometimes occurs in the middle of one table row. In the second case, multi-line text of that row is divided into two pages.  While texts of cells wrapping to become multi-line, the division occurs sometimes between the tokens and sometimes in a single token. When I extract texts of each cell using pdfplumber, currently the text comes as stripped from white spaces. So I should put a white space between parts while merging if the division occurred between tokens and I should not put if the division occurred in the middle of a single token. However, this is almost impossible to satisfy because there are also some cases with numbers. Therefore I need to get the texts of the cells as it is, with their leading and trailing white spaces, without stripping. 

According to my needs, I added this feature to the library, and tested on my task. I will commit and open a pull request. Feedback and improvements are welcome. 

Three examples about the issue and their explanations are below.

Example 1: 1st, 3rd, 6th and 7th columns need to be merged with a white space in between. 4th and 8th columns need to be merged directly. 
![Screenshot_20200927_172514](https://user-images.githubusercontent.com/50371163/94367467-af189000-00e7-11eb-8abd-937d71c14152.png)

Example 2: Numbers case: The number in 4th column should be merged directly without a white space.
![Screenshot_20200927_172921](https://user-images.githubusercontent.com/50371163/94367469-b049bd00-00e7-11eb-92e8-e817745cad37.png)

Example 3: Numbers case: The number in 7th column should be merged with a white space in between. 
![Screenshot_20200927_173202](https://user-images.githubusercontent.com/50371163/94367470-b049bd00-00e7-11eb-8bfc-afd9f742824e.png)

PR #278, opened 2020-09-27
feature: Option to not striping whitespaces from texts of cells.

AysegulKarcili:
For issue: Option to extract text of cells with leading and trailing white spaces preserved #277

### DIFF
--- a/README.md
+++ b/README.md
@@ -277,7 +277,7 @@ If you're using `pdfplumber` on a Debian-based system and encounter a `PolicyErr
 
 | Method | Description |
 |--------|-------------|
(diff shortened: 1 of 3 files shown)


samkit-jain (changes requested):
Thanks for the PR @AysegulKarcili. I don't imagine the PR breaking anything, but, could you please add a test case that demonstrates the need for this in which the test would pass on your branch but fail on the `jsvine:stable` branch? Please remove any sensitive information from the PDF that you are not comfortable sharing.

jsvine:
Yes, thanks for suggesting this @AysegulKarcili. Seems like a nice and simple feature. One other small request: Could you rename the parameter to `strip_whitespace` (without the plural `s`)?

PR #279, opened 2020-09-29
Allow GitHub Actions workflow to run on pull requests from forked repos

samkit-jain:
I noticed in https://github.com/jsvine/pdfplumber/pull/278 that the checks were stuck at "Waiting for status to be reported". This was happening because in our workflow file we had set it to run on all push commits and apparently, that is not enough for the workflow to run on forked repos. This PR attempts to solve that by adding a `pull_request_target` event trigger so that the workflow is run on PRs raised from forked repos. More information can be found [here](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request_target) and [here](https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,6 +1,6 @@
 name: Tests
 
-on: [push]
+on: [push, pull_request_target]
 
 jobs:
   lint:


samkit-jain:
The check is failing because of the newer black version. There are 2 options:
* Pin the black version so that the checks pass and since merging in `stable`, no conflicts when merging stable onto `develop`
* Reformat per the latest black version and change the base to `develop`

jsvine:
Thanks for this PR, @samkit-jain and for raising the question about `black`. I think we should pin `black` to a specific version. I've pushed a quick-fix to `stable` doing so, and will merge this PR.

Issue #71, opened 2018-07-31
Repeating characters


samkit-jain:
I'm facing a weird problem wherein characters are repeated when using `extract_text()` or `extract_tables()`. Example, `SSttaatteemmeenntt  ooff  AAccccoouunnttss` is printed instead of `Statement of Accounts`.

Sometimes, it happens in a portion of the PDF and sometimes in the whole PDF. When this happens in a portion of PDF, it is fixable (not completely) via `extract_text(x_tolerance=0, y_tolerance=0)` but not when the issue affects the whole PDF. Also, note that I do not face this issue in all PDFs but in some.

Lines are also repeated. Example,
```
Year-to-date totals do not reflect any fee or interest refunds
Year-to-date totals do not reflect any fee or interest refunds
you may have received.
you may have received.
```

samkit-jain:
On doing `first_page.extract_words(x_tolerance=0, y_tolerance=0)`, there are two instances of a single word
```
{'x0': Decimal('231.532'), 'x1': Decimal('252.251'), 'top': Decimal('916.343'), 'bottom': Decimal('925.422'), 'text': 'reflect'}
{'x0': Decimal('231.533'), 'x1': Decimal('252.252'), 'top': Decimal('916.383'), 'bottom': Decimal('925.462'), 'text': 'reflect'}
```

And repeating characters are still present for some words,
```
{'x0': Decimal('489.040'), 'x1': Decimal('506.160'), 'top': Decimal('269.320'), 'bottom': Decimal('277.480'), 'text': 'ttooddaayy'}
```

jsvine:
That's strange, indeed. My hunch is that there really are two copies of each letter in the PDF. (One set of letters might be transparent, perhaps?) What happens if you try extracting the text with another tool, such as `poppler-utils`'s `pdftotext`? (https://en.wikipedia.org/wiki/Pdftotext)

samkit-jain:
No such problem with `pdftotext`. This is the output,

**No repeating lines**
```
Year-to-date totals do not reflect any fee or interest refunds
you may have received.
```

**No repeating characters**
```
today
```
```
Statement of Accounts
```

jsfenfen:
[...]

PR #280, opened 2020-09-29
fix duplicates in extract_text/extract_words/extract_tables

xv44586:
https://github.com/jsvine/pdfplumber/issues/71


samkit-jain (commented):
Thanks for the PR @xv44586 Could you also please add a test case for the same?

xv44586:
> Thanks for the PR @xv44586 Could you also please add a test case for the same?

done! please check this out

jsvine:
Thank you, I will review this.

jsvine:
Many thanks for providing this inspiration and the test, @xv44586! I think that this proposed solution is slightly more complex than it needs to be; rather than adding a parameter to every relevant method, we can just provide a `utils` and `Page` method that allows users to deduplicate chars beforehand. In https://github.com/jsvine/pdfplumber/commit/04fd56ac405fd753e7f9c826ce103459013f2e71 I have added this logic (available in `develop` and in the next release), using your very helpful test and have credited you in the README. Thanks again 👍

Issue #292, opened 2020-10-20
Troubles with subscripts


Fleur09:
Hi, I want to extract some text with medical information, from pdf to txt files. 

The code I'm using to do it: 

import pdfplumber
import os

path = "PDF"
for filename in os.listdir(path):
     myFile = os.path.join(path, filename)
     with pdfplumber.open(myFile) as pdf:
          for page in pdf.pages:
               text = page.extract_text()
               with open(os.path.splitext(myFile)[0] + "_plumber.txt", encoding="utf-8", mode="a") as f:
                    f.write(text)

I need to keep the subscripts, not necessary as subscripts but in the text like "... CO2 ... ".
For now I have "CO" and the "2" is read like an other line like:
"..... CO .....
2"
As you can see:
![image](https://user-images.githubusercontent.com/56920799/96553333-4f13a480-12b5-11eb-8c3a-d4bef1239bf2.png)

One of the PDF  files: 
[doc_test3.pdf](https://github.com/jsvine/pdfplumber/files/5407798/doc_test3.pdf)

- pdfplumber version: I think I have the last one as I have installed it 2 weeks ago with "pip install".
- Python version: 3.8.6
- OS:  Windows 10

Thanks for your help.

samkit-jain:
Hi @Fleur09 Appreciate your interest in the library. You can make use of the `y_tolerance` param to solve for this. Use `page.extract_text(y_tolerance=6)` instead. The output would be 
```
Chapitre 5
Item 265
DÉSORDRES
DE L’ÉQUILIBRE ACIDE-BASE
UUEE  88..   CCiirrccuullaattiioonn  --  MMééttaabboolliissmmeess
NN°°226655..   TTrroouubblleess  ddee  ll’’ééqquuiilliibbrree  aacciiddoo--bbaassiiqquuee  eett  ddééssoorrddrreess  hhyyddrroo--éélleeccttrroollyyttiiqquueess..
OOBBJJEECCTTIIFFSS
yy  PPrreessccrriirree  eett  iinntteerrpprréétteerr  uunn  eexxaammeenn  ddeess  ggaazz  dduu  ssaanngg  eett  uunn  iioonnooggrraammmmee  ssaanngguuiinn  eenn  
ffoonnccttiioonn  dd’’uunnee  ssiittuuaattiioonn  cclliinniiqquuee  ddoonnnnééee..
[...]

PR #294, opened 2020-10-20
Fix bug in `dedup_chars()` in which `._objects` was accessed before assignment

### LINKED ISSUE #293: AttributeError: 'Page' object has no attribute '_objects'
## Describe the bug

Instance variable `_objects` in class `Page` accessed before assignment in `dedupe_chars()`. `_objects` gets instantiated only when the property method `objects` of the `Page` class is invoked. Source: https://github.com/jsvine/pdfplumber/blob/04fd56ac405fd753e7f9c826ce103459013f2e71/pdfplumber/page.py#L148-L153

```python
import pdfplumber
pdf = pdfplumber.open("examples/pdfs/ag-energy-round-up-2017-02-24.pdf")
page = pdf.pages[0]
page.dedupe_chars()
```

## PDF file

https://github.com/jsvine/pdfplumber/blob/stable/examples/pdfs/ag-energy-round-up-2017-02-24.pdf

## Expected behavior

No error.

## Actual behavior

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/samkitjain/Samkit/Projects/Python/pdfplumber/pdfplumber/page.py", line 273, in dedupe_chars
    p._objects = dict((kind, objs) for kind, objs in self._objects.items())
AttributeError: 'Page' object has no attribute '_objects'
```

## Environment

- pdfplumber version: `develop` branch
- Python version: 3.8.2
- OS: Ubuntu 18.04

samkit-jain:
This PR fixes #293. Was uncaught in https://github.com/jsvine/pdfplumber/commit/04fd56ac405fd753e7f9c826ce103459013f2e71 probably because other methods may have called `.objects` before the call to `.dedup_chars()`.

To resolve the issue, I replaced the usage of `._objects` with `.objects`. If only `._objects` is to be used, can add a simple `self.objects` statement like
```python
    def dedupe_chars(self, **kwargs):
        """
        Removes duplicate chars — those sharing the same text, fontname, size,
        and positioning (within `tolerance`) as other characters on the page.
        """
        self.objects  # Statement has no effect but is useful as it would instantiate `._objects` for further use below
        p = FilteredPage(self, True)
        p._objects = dict((kind, objs) for kind, objs in self._objects.items())
        p._objects["char"] = utils.dedupe_chars(self.chars, **kwargs)
        return p
```
Can also do the same in the `__init__()` method instead as well.

NOTE: I added a new PDF `issue-71-duplicate-chars-2.pdf` (thanks to @Fleur09 for sharing in #292) as it has repeating characters.


jsvine (approved):

jsvine:
Great catch, thanks! Merging.

Issue #297, opened 2020-10-27
Decode Integer Metadata


prgx-csmith01:
I have received this error message for a PDF file:

```
Traceback (most recent call last):
  File "####", line 145, in ####
    pdf = pdfplumber.load( #### )
  File "/opt/venv-app-microservice/lib/python3.7/site-packages/pdfplumber/__init__.py", line 11, in load
    return PDF(file_or_buffer, **kwargs)
  File "/opt/venv-app-microservice/lib/python3.7/site-packages/pdfplumber/pdf.py", line 42, in __init__
    self.metadata[k] = decode_text(v)
  File "/opt/venv-app-microservice/lib/python3.7/site-packages/pdfplumber/utils.py", line 70, in decode_text
    ords = (ord(c) if type(c) == str else c for c in s)
TypeError: 'int' object is not iterable
```

It seems that there is no handling for integer metadata in the __init__ of pdf.py

Previously there was a similar bug raised [#67](https://github.com/jsvine/pdfplumber/issues/67) for boolean objects.

I cannot provide the PDF used that caused this error as it is client data. The metadata of the file contains { ... , "Copies" : 0 }.

samkit-jain:
Hi @prgx-csmith01 Would it be possible for you to redact everything from the PDF and then share it so that it can be added as a test to PR #298 ?

prgx-csmith01:
Hi @samkit-jain , I can't share the PDF but we have created a test file for you with an example of the metadata issue. I hope this helps. Thanks!

[test_int_metadata.pdf](https://github.com/jsvine/pdfplumber/files/5457629/test_int_metadata.pdf)

samkit-jain:
Many thanks @prgx-csmith01 I have updated the PR #298 with the test case.

mkl-public:
As an aside: That integer value of the **Copies** entry is invalid.

According to the specification:

> ### 14.3.3 Document Information Dictionary
>
> ...
>
>The value associated with any key not specifically mentioned in Table 317 shall be a text string.

_(ISO 32000-1)_

... and neither is there any **Copies** entry in table 317 nor any other entry with a numeric type, merely text strings, dates, and names.

Thus, this issue strictly speaking is not a _bug_ (as labeled currently) but a request to support one more type of invalid PDFs.

jsvine:
[...]

PR #298, opened 2020-10-27
Handle integer/floating type metadata values

samkit-jain:
This is a quick fix for issue #297 

A test is yet to be added as I have been unable to edit a metadata value and store it as an integer type. It keeps getting saved as a string. If any of the watchers of this PR can modify the metadata of any of the available PDFs in the repo such that the data type becomes an integer or a float, it would be much appreciated and I'll add a test case for it as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to this project will be documented in this file. The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
+## [0.5.25] — Unreleased
+### Fixed
+- Fix metadata extraction to handle integer/floating-point values ([#297](https://github.com/jsvine/pdfplumber/issues/297))
+
 ## [0.5.24] — 2020-10-20
 ### Added
 - Added `extra_attrs=[...]` parameter to `.extract_text(...)` ([c8b200e](https://github.com/jsvine/pdfplumber/commit/c8b200e)) ([#28](https://github.com/jsvine/pdfplumber/issues/28))
--- a/pdfplumber/pdf.py
+++ b/pdfplumber/pdf.py
@@ -33,10 +33,10 @@ class PDF(Container):
                 self.metadata[k] = list(map(decode_text, v))
             elif isinstance(v, PSLiteral):
                 self.metadata[k] = decode_text(v.name)
-            elif isinstance(v, bool):
-                self.metadata[k] = v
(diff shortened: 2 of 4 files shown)


jsvine:
Many thanks, @samkit-jain! And, yes, if anyone has a PDF that could serve as a test case, that would be great. In the meantime, I’ll see if I can find one.

samkit-jain:
@jsvine Updated the PR with the test case. Also acknowledged @prgx-csmith01 in the commit message.

pdfplumber/pdf.py
```diff
@@ -33,10 +33,10 @@ def __init__(self, stream, pages=None, laparams=None, precision=0.001, password=
                 self.metadata[k] = list(map(decode_text, v))
             elif isinstance(v, PSLiteral):
                 self.metadata[k] = decode_text(v.name)
-            elif isinstance(v, bool):
-                self.metadata[k] = v
-            else:
+            elif isinstance(v, str) or isinstance(v, bytes):
```
jsvine: I think this can be slightly simpler: `elif isinstance(v, (str, bytes)):`

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to this project will be documented in this file. The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
+## [0.5.25] — Unreleased
+### Fixed
+- Fix metadata extraction to correctly handle integer/floating-point values ([#297](https://github.com/jsvine/pdfplumber/issues/297))
```
jsvine: Given the spec, [as recently noted in the issue thread](https://github.com/jsvine/pdfplumber/issues/297#issuecomment-718862330), perhaps it would be more accurate to remove the word "correctly"?

jsvine (changes requested):
Thanks, @samkit-jain! Just a couple small notes. Really appreciate you putting together the test for this.

pdfplumber/pdf.py
samkit-jain: Thanks for sharing this @jsvine I didn't know you could specify a tuple in `isinstance()`. Learnt something new today :)

CHANGELOG.md
samkit-jain: Sure

jsvine (approved):
Thanks, looks great! Merging.

PR #302, opened 2020-11-01
Add support fill value to all the cells belongs to the same merged cell

### LINKED ISSUE #79: Dealing with merged table cells
I'm doing table extraction on a PDF generated by Word, and there are many merged cells in the table.

Now if I use `extract_table`, I would get a table like this, with many None cells.

I want to praise this tool for its excellent accuracy on OCR! And I would be even more grateful if there is a new, easy function that can deal with these merged cells properly.

For example, I would expect the 3rd column to be also “指标”, cell at loc[0,"项目"] to be also "项目", and cell at loc[2,None] to be also 8.0, etc. To achieve this now, I must use complex `fillna` logics considering the positions of `cells`.

Thanks a lot!

guo1017138:
If tables contains merged cell, the value will fill to left up cell only,
others in the same merged cell will be empty. With new option
merged_cell_fullfill set to True in table settings, all the cells will
have the same value. Fix #79

Signed-off-by: Guo Larry <guo1017138@163.com>


jsvine:
Hi @guo1017138, and thanks for your interest in this library. Without any tests or examples, it's hard to determine whether this code achieves its goal, whether this part of the library is the best place to put such code, or whether this is the optimal approach more generally. If you provide more details and tests, we can discuss those topics.

guo1017138:
@jsvine Thanks for the comments. I've add the examples and test cases for verifying it.
Please help to review.

jsvine:
Thank you for sharing this @guo1017138. I now have a much better understanding of the goal of this PR. I appreciate it. 

Unfortunately, I don't think this PR/addition a good fit for this library at this time. One reason is that the functionality the PR adds can be achieved by the user *outside* of the extraction code, without adding the additional complexity to the existing codebase. Additionally, I am contemplating future modifications to the output of `.extract(...)` that would handle merged/nested cells in a clearer way than the library currently does, so I don't want to promise users that `merged_cell_fullfill` would continue to work. 

That said: Thank you again. I'm glad to see your approach to handling merged cells and I will take it into consideration as I continue work on the library.

guo1017138:
@jsvine , Thanks. That's OK. I will use my private code until this library have official handling for merged cells.

Issue #316, opened 2020-11-23
TypeError: 'PDFObjRef' object is not iterable


Abdur-rahmaanJ:
## Describe the bug

Got 

```bash
Traceback (most recent call last):
  File "main.py", line 5, in <module>
    with pdfplumber.open("<stripped_path>") as pdf:
  File "<stripped_path>\venv\lib\site-packages\pdfplumber\pdf.py", line 46, in open
    return cls(open(path_or_fp, "rb"), **kwargs)
  File "<stripped_path>\venv\lib\site-packages\pdfplumber\pdf.py", line 33, in __init__
    self.metadata[k] = list(map(decode_text, v))
  File "<stripped_path>\pdfres\venv\lib\site-packages\pdfplumber\utils.py", line 77, in decode_text
    ords = (ord(c) if type(c) == str else c for c in s)
TypeError: 'PDFObjRef' object is not iterable
```

## Code to reproduce the problem
```python
import pdfplumber

with pdfplumber.open("target.pdf") as pdf:
    first_page = pdf.pages[0]
    print(first_page.chars[0])
```

## Environment

- pdfplumber version: 0.5.24
- Python version: 3.8.1
- OS: Windows

samkit-jain:
Hi @Abdur-rahmaanJ Appreciate your interest in the library. Would it be possible for you to share a PDF to demonstrate this issue? Will help us in reproducing and fixing the issue. Please remove any sensitive information from the PDF before sharing it here.

Abdur-rahmaanJ:
Try it on any research gate pdf. If you dont get the error on windows, i'll send you the exact pdf

samkit-jain:
I chose [this](https://www.researchgate.net/publication/344552831_The_Role_of_Race_and_Gender_in_Teaching_Evaluation_of_Computer_Science_Professors_A_Large_Scale_Analysis_on_RateMyProfessor_Data) PDF and it ran fine for me. I got
```
[...]

PR #320, opened 2020-11-29
Handle invalid metadata values

samkit-jain:
This PR fixes issue #316 

**Background:** The PDF exhibiting the exception contained a metadata key `Changes` which was a list of `PDFObjRef` objects. The code responsible for resolving metadata expected a list to be of string-like objects and since in this case, it wasn't, the `decode_text()` function raised the `TypeError` exception as mentioned in the issue.

This metadata value type is not as per the [PDF specification (see p556 sec14.3)](https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf). PDF specifications are well, just specifications. Not conforming to it does not necessarily render a PDF unreadable and a lot of PDFs don't abide by every single rule mentioned in it. Hence, in the first commit, I added a `try-except` block that would log a warning to the console when a metadata value is unparseable and continues to the next one. This would allow the user to perform non-metadata related operations without any issue. I think we can also make this behaviour configurable by adding a new parameter, say, `fail_on_metadata` in the `PDF.open()` method that when set to `True`, will raise an exception and when `False`, just log to the console.

**Fix:** To resolve the issue, I have added a new method `resolve_and_decode()` which would recursively go through the metadata values and decode them.

**Additional:** Pinned versions of `pytest` and removed unused imports in `tests/test_issues.py`.


jsvine:
Thanks, @samkit-jain! I think this is a smart approach. Just a few notes / requests:

- > I think we can also make this behaviour configurable by adding a new parameter, say, `fail_on_metadata` in the `PDF.open()` method that when set to `True`, will raise an exception and when `False`, just log to the console. 
  - I like that. Want to try implementing on this PR? Perhaps the parameter, however, should be `strict_metadata`? Open to other ideas.

- Would you be able to add a test that brings the coverage for `pdf.py` back up to 100%?

- Can you add the related entries to the `CHANGELOG.md`?

Thanks again!

samkit-jain:
@jsvine I have made the metadata failure condition configurable. Since I am unable to create a PDF that would cause an exception in the new recursive resolver method, I was unable to increase the overall test coverage. Added tests for the draw methods to increase the coverage instead, but, this meant that the test coverage for the files changed in this PR is still below the previous one. That's why the [codecov/patch](https://github.com/jsvine/pdfplumber/pull/320/checks?check_run_id=1518890701) is failing but [codecov/project](https://github.com/jsvine/pdfplumber/pull/320/checks?check_run_id=1518890695) is passing. The changelog has been updated as well.

jsvine:
Thanks, @samkit-jain! I'm slightly wary of incorporating new code that doesn't have a PDF to test it with, but have merged this since (a) the logic of the code makes sense, (b) the untested part of the code handles what would already be edge-cases, and (c) the tests pass for all the PDFs we *are* using.

PR #338, opened 2021-01-19
Sort tables extracted on a page by their `top` position

### LINKED ISSUE #336: tables in a page getting wrong order
Hello,

First thank you for this great library, it really help me to quick process pdf files.

When i use **page.extract_tables** to extract tables in a same page, i get tables in a wrong order.

Why this problem occurs?

Thanks a lot for answer me

samkit-jain:
Fixes #336
h/t @gqh1995 for reporting

When multiple tables are found on a page, sorts them top to bottom. If there is a collision, sorts left to right. Not sure, but I think instead of sorting as I have done here, but perhaps, can update the sorting at https://github.com/jsvine/pdfplumber/blob/de767c3a52bbcacd237a17a051a3975d592023d9/pdfplumber/table.py#L315 instead.

PS: The changelog would need to be updated with the commit hash after this PR gets merged in.


CHANGELOG.md
```diff
@@ -2,8 +2,13 @@
 
 All notable changes to this project will be documented in this file. The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
+## [0.5.26] — TBD
+### Fixed
+- Fix ordering of tables to be top to bottom ([#336](https://github.com/jsvine/pdfplumber/issues/336))
```
jsvine: Given that this was not an explicit bug, but rather a change/improvement, let's move this to a `### Changed` header and rephrase to: "Change `TableFinder` to return tables in order of topmost-and-then-leftmost, rather than leftmost-and-then-topmost"

pdfplumber/table.py
```diff
@@ -450,6 +450,11 @@ def __init__(self, page, settings={}):
         )
         self.cells = intersections_to_cells(self.intersections)
         self.tables = [Table(self.page, t) for t in cells_to_tables(self.cells)]
+        # If multiple tables are found, sort them by their vertical placement.
+        # Top to bottom, left to right.
+        self.tables = sorted(
```
jsvine: I'd rather not assign `self.tables` twice. And, given that we're just sorting a list returned by `cells_to_tables(...)`, I think the sorting probably fits best in that function. 

In fact, there's already a line there that handles sorting: https://github.com/jsvine/pdfplumber/blob/de767c3a52bbcacd237a17a051a3975d592023d9/pdfplumber/table.py#L315

That logic identifies each table's leftmost-and-then-topmost cell (due to the `(left, top, right, bottom)` sequence of the bbox tuples), and then does the same comparison between tables. I think you're right, though, that topmost-and-then-leftmost is a more reasonable expectation, especially when the left-alignment of a series of tables might differ by just a few (otherwise imperceptible) decimal places.

jsvine (changes requested):
Thanks so much for this, @samkit-jain. I think you've identified a helpful improvement. It may cause problems for some (hopefully very few) people — if they depend on the current leftmost-and-then-topmost ordering and have a PDF where the left-alignment is either meaningful or predictably uneven — but I think it will save people more trouble in the long run.

Many thanks for adding the test, too. I've made a suggestion about moving the logic to a slightly different place in the code, but this otherwise looks good.

samkit-jain:
Thanks for the review @jsvine How about we add a new argument `table_direction` that controls whether the tables are read leftmost (default) or topmost? This would ensure that the change does not break any integrations.

jsvine:
My initial instinct is not to add this argument. My hunch is that this change will affect very few integrations, and that there is (and will remain) a fairly easy way to customize the table order, by using `page.find_tables(...)` instead of `page.extract_tables(...)`. E.g.,:

```python
tables = sorted(my_page.find_tables(), key=lambda x: x.bbox)
extracted = [ t.extract() for t in tables ]
```

That said, I'm open to being persuaded otherwise. Thoughts?

pdfplumber/table.py
samkit-jain: Yeah, I wasn't sure what the `min(t["corners"])` was doing. Thanks for explaining. Here's what I have done: Added `[::-1]` to it so that the tuple gets reversed to become `(y, x)` instead of `(x, y)`. Removed the additional `sorted(...)` call too. Added a comment as well that explains it. I hope that's fine.

samkit-jain:
@jsvine I agree. The user has control over the sorting using the `.find_tables(...)` method which makes adding a new argument redundant.

pdfplumber/table.py
```diff
@@ -312,7 +312,11 @@ def assign_cell(cell, table):
     if len(current_table["cells"]):
         tables.append(current_table)
 
-    _sorted = sorted(tables, key=lambda t: min(t["corners"]))
+    # Sort the tables identified in order of topmost-and-then-leftmost by sorting via
+    # the Y axis. ``min(t["corners"])`` would give the (X, Y) coordinate of the table's
+    # topleft corner. ``[::-1]`` reverses the tuple so that the sorting happens first on
+    # the Y axis and then on X.
+    _sorted = sorted(tables, key=lambda t: min(t["corners"])[::-1])
```
jsvine: My understanding of this line is that it would do this:

- Find the leftmost-and-then-topmost vertex/corner of each table — `min(t["corners"])`
- Reverse the tuple representation of that specific vertex for each table, so that it is `y,x` instead of `x,y`
- Sort by those reversed tuples

Instead, how about this:

```python
_sorted = sorted(tables, key=lambda t: min(c[::-1] for c in t["corners"]))
```

... which does the reversal before finding the `min` corner? Or perhaps I'm mixed up?

samkit-jain: Your suggestion is also correct @jsvine but the top left corner of a rectangle will have both the minimum x and y values and hence, both `min(t["corners"])[::-1]` and `min(c[::-1] for c in t["corners"])` should give the same final output provided the rectangle figure in question is upright.
image
I didn't consider rotated/skewed rectangles though. Should I?

jsvine: My apologies; I should have described my reasoning more explicitly here. I was thinking less about skewed rectangles, and more about uneven tables, e.g.,:

```
    +---+---+---+
    | A | B | C |
+---+---+---+---+
| E | F | G | H |
+---+---+---+---+
| I | J | K | L |
+---+---+---+---+
```

... where the top-left corner of `A` would be the `topmost-and-then-leftmost` corner, but the top-left corner of `E` would be the `leftmost-and-then-topmost` corner. I admit that this probably a rare case where this would matter ... but it does seem possible.

samkit-jain: Thanks for the example. The suggestion makes more sense now. Made the change.

jsvine (approved):
Great, thanks! Will merge.

PR #70, opened 2018-07-18
reset wand alpha_channel to backgound white

meldonization:
in some cases, wand image alpha channel could make the whole page black out. -_-||

- raw pdf:
[background_alpha_channel.pdf](https://github.com/jsvine/pdfplumber/files/2204779/background_alpha_channel.pdf)

-converted image w/o background white out
background_alpha_channel

- converted image w/ background white out
whiteout_background

PR #340, opened 2021-01-23
Rework the alpha conversion to remove jaggies

arlyon:
The current method doesn't antialias correctly, presumably because the alpha layer is discarded while there is still data. This new method uses wand to handle the background, allowing us also to remove the special case conversion logic.

Also, am I right in saying that  the docstring here is inaccurate? I don't see the kwargs being passed in anywhere. If we were, for example, to support setting the `background` arg, we'd have to support the alpha channel, as transparent is valid. Happy to update the docstring once I get some more context! :)

### DIFF
--- a/pdfplumber/display.py
+++ b/pdfplumber/display.py
@@ -21,10 +21,6 @@ DEFAULT_RESOLUTION = 72
 
 
 def get_page_image(stream, page_no, resolution):
-    """
-    For kwargs, see http://docs.wand-py.org/en/latest/wand/image.html#wand.image.Image
-    """
-
     # If we are working with a file object saved to disk
     if hasattr(stream, "name"):
         spec = dict(filename=f"{stream.name}[{page_no}]")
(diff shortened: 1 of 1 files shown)


arlyon:
To compare the difference in output, see below (large images, you may need to zoom):

image

image

jsvine:
Thanks, @arlyon! This seems like a nice improvement. (Note to my future self: The initial setting of "background" came via https://github.com/jsvine/pdfplumber/pull/70.) 

Re. this:

> Also, am I right in saying that the docstring here is inaccurate? I don't see the kwargs being passed in anywhere. If we were, for example, to support setting the `background` arg, we'd have to support the alpha channel, as transparent is valid. Happy to update the docstring once I get some more context! :)

... it appear that you are correct; the docstring is inaccurate. It has been around since 0.5.0, introduced via https://github.com/jsvine/pdfplumber/pull/23; I likely started with a `**kwargs`-based parameter list but scrapped it without updating the docstring. Thank you for catching that.

It sounds like you're more familiar with Wand and image processing than I am. Do you see a particular advantage in users being able to pass `**kwargs` to this method? And where would they be applied? (To `wand.image.Image(...)`?)

And one request: Could you change the target of this PR from `stable` to `develop`?

arlyon:
There's probably not much of interest except maybe the background colour but if someone want their documents in pink they can open an issue :) haha

I've just deleted the docstring for now, I felt anything I tried to write was fairly self explanatory from the code itself, and I've rebased!

Thanks for the quick reply

jsvine (approved):
Looks good and makes sense. Thanks, and merging!

Issue #362, opened 2021-02-26
Error importing in ubuntu


ghost:
Hello! When i´m trying importing in ubuntu, this error is displayed:

```
>>> import pdfplumber
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.5/dist-packages/pdfplumber/__init__.py", line 10, in <module>
    from .pdf import PDF
  File "/usr/local/lib/python3.5/dist-packages/pdfplumber/pdf.py", line 49
    f'[WARNING] Metadata key "{k}" could not be parsed due to '
                                                              ^
SyntaxError: invalid syntax

```

jsvine:
My best guess is that your environment (or virtual environment) is using Python 2, rather than Python 3.

samkit-jain:
>  File "/usr/local/lib/python3.5/dist-packages/pdfplumber/pdf.py", line 49

@betances It seems like you are using Python 3.5. The support for Python 3.5 was dropped in [version 0.5.23](https://github.com/jsvine/pdfplumber/blob/stable/CHANGELOG.md#0523--2020-08-15). Request you to upgrade to Python 3.6 or higher if you want to use pdfplumber 0.5.23 or higher else downgrade to pdfplumber 0.5.22 or lower.

jsvine:
Ah, good catch, @samkit-jain, and thanks for reading the traceback more carefully than I did! I concur.

ghost:
> > File "/usr/local/lib/python3.5/dist-packages/pdfplumber/pdf.py", line 49
> 
> @betances It seems like you are using Python 3.5. The support for Python 3.5 was dropped in [version 0.5.23](https://github.com/jsvine/pdfplumber/blob/stable/CHANGELOG.md#0523--2020-08-15). Request you to upgrade to Python 3.6 or higher if you want to use pdfplumber 0.5.23 or higher else downgrade to pdfplumber 0.5.22 or lower.

Thanks! :)

PR #363, opened 2021-02-26
Prevent pip from installing the project on Python versions 3.5 and lower

samkit-jain:
Result of #362, this PR enforces the Python version to be a minimum of 3.6 for pip to install the package.

Did some black formatting along the way as well.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,17 +1,19 @@
-import sys, os
+import os
+
 from setuptools import setup, find_packages
-import subprocess
 
 NAME = "pdfplumber"
 HERE = os.path.abspath(os.path.dirname(__file__))
 
 version_ns = {}
 
+
 def _open(subpath):
     path = os.path.join(HERE, subpath)
     return open(path, encoding="utf-8")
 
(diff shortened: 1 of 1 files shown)


samkit-jain:
I wasn't sure if the PR required the changelog to be updated. @jsvine How do you build the package before publishing it to pip? I think we can include that as a build verification step to the GitHub Actions workflow.

jsvine (approved):

jsvine:
> Result of #362, this PR enforces the Python version to be a minimum of 3.6 for pip to install the package.

Thanks, and good idea! Merging.

> I wasn't sure if the PR required the changelog to be updated.

I don't think it does, since we've already made clear that 3.5 support is deprecated, but thanks for asking.

> @jsvine How do you build the package before publishing it to pip? I think we can include that as a build verification step to the GitHub Actions workflow.

`python setup.py build sdist`

Feel free to open a PR with that, or I can add it to my todo list.

samkit-jain:
>Feel free to open a PR with that, or I can add it to my todo list.

Opened PR #365

PR #385, opened 2021-03-19
Fixed small issue with ordering of statements in `extract_text`

alexreg:
`chars = to_list(chars)` and `len(chars)` need to be swapped to avoid error sometimes.

Fixes issue #386.

### DIFF
--- a/pdfplumber/table.py
+++ b/pdfplumber/table.py
@@ -389,7 +389,9 @@ class Table(object):
 
                     if len(cell_chars):
                         cell_text = utils.extract_text(
-                            cell_chars, x_tolerance=x_tolerance, y_tolerance=y_tolerance
+                            cell_chars,
+                            x_tolerance=x_tolerance,
+                            y_tolerance=y_tolerance,
                         ).strip()
                     else:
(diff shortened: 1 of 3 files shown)


samkit-jain (changes requested):
Appreciate you opening a PR @alexreg Request you to add a test case that passes on this commit and fails otherwise.

alexreg:
@samkit-jain Sure, will do. How do I run tests though? Ideally, this should be documented in the README.

jsvine:
Thanks, @alexreg and @samkit-jain. I really appreciate the test addition — generally ... but also specifically in this case, because it clarifies what's going on here: The previous code didn't work when the `chars` argument is a generator. It might be a fairly edge-case situation, but since the fix is very straightforward and doesn't affect performance, I'll merge it.

jsvine:
And, yes, @alexreg — it's definitely worth us adding PR guidelines (including re. tests). Will open an issue on that and self-assign.

alexreg:
@jsvine That was my thought too, basically. Thanks for merging.

Issue #383, opened 2021-03-18
`KeyError` raised when `laparams` set


alexreg:
## Describe the bug

I get the following exception when calling `page.objects["char"]`. Note, this only occurs when I open the PDF with `laparams` set.

```
  File "foo.py", line 30
    chars = left_column.objects["char"]
  File "/usr/local/lib/python3.9/site-packages/pdfplumber/page.py", line 358, in objects
    self._objects = self.crop_fn(self.parent_page.objects, self.bbox)
  File "/usr/local/lib/python3.9/site-packages/pdfplumber/page.py", line 358, in objects
    self._objects = self.crop_fn(self.parent_page.objects, self.bbox)
  File "/usr/local/lib/python3.9/site-packages/pdfplumber/utils.py", line 477, in crop_to_bbox
    return dict((k, crop_to_bbox(v, bbox)) for k, v in objs.items())
  File "/usr/local/lib/python3.9/site-packages/pdfplumber/utils.py", line 477, in <genexpr>
    return dict((k, crop_to_bbox(v, bbox)) for k, v in objs.items())
  File "/usr/local/lib/python3.9/site-packages/pdfplumber/utils.py", line 481, in crop_to_bbox
    cropped = list(filter(None, (clip_obj(obj, bbox) for obj in objs)))
  File "/usr/local/lib/python3.9/site-packages/pdfplumber/utils.py", line 481, in <genexpr>
    cropped = list(filter(None, (clip_obj(obj, bbox) for obj in objs)))
  File "/usr/local/lib/python3.9/site-packages/pdfplumber/utils.py", line 422, in clip_obj
    overlap = get_bbox_overlap(obj_to_bbox(obj), bbox)
KeyError: 'x0'
```

## Code to reproduce the problem

```python3
with pdfplumber.open("serials.pdf", laparams = {}) as pdf:
	for page in pdf.pages:
		contents = page.crop(
			(
				Decimal(100),
				Decimal(70 + 200 if page.page_number == 1 else 0),
				page.width - Decimal(100),
				page.height - Decimal(70),
			),
		)
		left_column = contents.crop(
			(
				Decimal(0),
				Decimal(0),
				contents.width * Decimal(0.5),
				contents.height,
			),
			relative = True,
		)
		right_column = contents.crop(
			(
				contents.width * Decimal(0.5),
				Decimal(0),
				contents.width,
				contents.height,
			),
			relative = True,
		)

		chars = left_column.objects["char"]
```

## PDF file

[...]

PR #388, opened 2021-03-19
Even if laparams is set, don't extract `anno` objs

jsvine:
pdfminer.six's `LTAnno` objects are not PDF annotations (which we already provide access to via `.annots`, regardless of whether `laparams` is set), but rather layout annotations. Per [pdfminer.six codebase](https://github.com/pdfminer/pdfminer.six/blob/ef4787d8add35c84ed324e488a0683485b422f45/pdfminer/layout.py#L264-L270):

> Note that, while a LTChar object has actual boundaries, LTAnno objects does not, as these are "virtual" characters, inserted by a layout analyzer according to the relationship between two characters (e.g. a space).

Because they have no boundaries, they cause problems for pdfplumber, which expects bounding-box coordinates for all objects. See, e.g., issue #383, which this commit should fix.

### DIFF
--- a/pdfplumber/page.py
+++ b/pdfplumber/page.py
@@ -221,6 +221,8 @@ class Page(Container):
         objects = {}
         for obj in self.iter_layout_objects(self.layout._objs):
             kind = obj["object_type"]
+            if kind in ["anno"]:
+                continue
             if objects.get(kind) is None:
                 objects[kind] = []
             objects[kind].append(obj)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file. The format 
 ## [0.5.28] — Unreleased
 ### Fixed
 - Fix `.extract_text(...)` so that it can accept generator objects as its main parameter. ([#385](https://github.com/jsvine/pdfplumber/pull/385)) [h/t @alexreg]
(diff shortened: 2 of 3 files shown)


pdfplumber/page.py
```diff
@@ -221,6 +221,8 @@ def parse_objects(self):
         objects = {}
         for obj in self.iter_layout_objects(self.layout._objs):
             kind = obj["object_type"]
+            if kind in ["anno"]:
```
samkit-jain, line 224: Why not `kind == "anno"` and why `kind in ["anno"]`?

jsvine: My thinking here was that other `pdfminer.six` object types might need to be excluded in the future, and to very subtly (perhaps too subtly) indicate that through the in-array syntax. I can see the argument for making the comparison more direct, though. What do you think?

samkit-jain: I guess it is fine if we can expect new keys to be added here in the near future. Approving the PR. Thanks.

samkit-jain (approved):

samkit-jain:
@jsvine I noticed a check waiting to be reported as in the screenshot
image
You would need to update the required checks in the branch protection settings for it to disappear.

jsvine:
Thanks! And strange about the still-pending check.

PR #407, opened 2021-04-08
Add --laparams to CLI (and make related tweaks)

jsvine:
This commit adds an `--laparams` flag to the pdfplumber CLI, giving it more feature parity with the core library. To do so, it makes some internal changes to `convert.py`, including changing the list of objects to convert from *a predefined default list* to *all types extracted*. It also changes the order of `convert_csv`'s output so that objects are sorted first by page number (rather than object type).

### DIFF
--- a/pdfplumber/cli.py
+++ b/pdfplumber/cli.py
@@ -3,6 +3,7 @@ from . import convert
 from .pdf import PDF
 import argparse
 from itertools import chain
+import json
 import sys
 
 
@@ -23,12 +24,15 @@ def parse_args(args_raw):
 
     parser.add_argument("--format", choices=["csv", "json"], default="csv")
 
+    parser.add_argument("--types", nargs="+")
+
     parser.add_argument(
-        "--types",
-        nargs="+",
-        default=convert.DEFAULT_TYPES,
-        choices=convert.DEFAULT_TYPES,
+        "--all-types",
+        action="store_true",
+        help="Return all types of objects. Overrides --types.",
     )
+
+    parser.add_argument("--laparams", type=json.loads)
 
     parser.add_argument("--pages", nargs="+", type=parse_page_spec)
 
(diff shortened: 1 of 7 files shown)


pdfplumber/container.py
```diff
@@ -47,6 +47,22 @@ def images(self):
     def chars(self):
         return self.objects.get("char", [])
 
+    @property
+    def textboxverticals(self):
+        return self.objects.get("textboxvertical", [])
```
samkit-jain, line 52: The name of the method is plural but the name of the key is singular. Shouldn't both be same?

pdfplumber/cli.py
```diff
@@ -23,13 +24,16 @@ def parse_args(args_raw):
...
     parser.add_argument("--format", choices=["csv", "json"], default="csv")
 
+    parser.add_argument("--types", nargs="+")
+
     parser.add_argument(
-        "--types",
-        nargs="+",
-        default=convert.DEFAULT_TYPES,
-        choices=convert.DEFAULT_TYPES,
+        "--all-types",
```
samkit-jain: Where is this parameter being used?

jsvine: Whoops! Nowhere. It's a vestige of a previous approach I considered, but decided not to pursue. Now removed in latest commit. Thanks for catching.

pdfplumber/container.py
jsvine: I also find this unsatisfying, but it parallels the approach of the preexisting code — e.g., `page.objects["char"]` + `page.chars`. I'm open to altering this approach in a future major version. Or perhaps I'm overlooking a more elegant solution? Thoughts?

samkit-jain: I don't have any answer at the moment but yeah, something to consider changing in a next major version.

samkit-jain (approved):

PR #415, opened 2021-04-17
Pass kwds args to Table.extract from Page.extract_table

trifling:
The extract_table method of the Page class does not honour text tolerance settings. Pass the relevant parameters from the table_settings parameter to the Table.extract method.

### DIFF
--- a/pdfplumber/page.py
+++ b/pdfplumber/page.py
@@ -256,7 +256,14 @@ class Page(Container):
             return (-len(x.cells), x.bbox[1], x.bbox[0])
 
         largest = list(sorted(tables, key=sorter))[0]
-        return largest.extract()
+
+        extract_kwargs = dict(
+            (k, table_settings["text_" + k])
+            for k in ["x_tolerance", "y_tolerance"]
+            if "text_" + k in table_settings
+        )
+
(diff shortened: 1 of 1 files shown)


samkit-jain (commented):
Hi @trifling, appreciate you raising the PR. It seems like a valid bug fix to me if you compare with the same operation being done in the [`.extract_tables(...)`](https://github.com/trifling/pdfplumber/blob/stable/pdfplumber/page.py#L240-L246) above. Nice catch!

Request you to do the following
1. Add a test case that demonstrates this fix. The test should pass with the changes in this PR but fail on the commits prior.
2. Switch the base from `stable` to `develop`.
3. Add yourself to the [list of contributors](https://github.com/jsvine/pdfplumber#acknowledgments--contributors).
4. Update the [changelog](https://github.com/jsvine/pdfplumber/blob/stable/CHANGELOG.md).

jsvine:
Given the simplicity of this change and a desire to get this into `0.5.28`, I've merged this, added the test, and updated the contributor list and changelog. Thanks for catching this, @trifling, and  thanks for flagging the necessary additions, @samkit-jain.

Issue #420, opened 2021-04-22
Duplicate value for merged cell instead of `None`


tungph:
Hi there,
Firstly, I want to thank you for creating this awesome `pdfplumber` project. I have been using it a lot lately. 
Secondly, I'm facing a problem and want to propose change to the output of the table module regarding the utility `extract_tables` with merged cells.
As it is now, only the first row of the merged cell has value, the remaining cells get `None`.
![image](https://user-images.githubusercontent.com/4607414/115679016-54dd6600-a37c-11eb-99ae-102bfe91c41f.png)

```
Header 1, Header 2
Merged cell 1, Cell 2-1
None, Cell 2-2
```

It would be more semantically correct to have the value in merged cell duplicated like this
```
Header 1, Header 2
Merged cell 1, Cell 2-1
Merged cell 1, Cell 2-2
```

I'm preparing a PR, please consider it.
Thank you!

### Update 1
Here's the PR: #422

Approach: finding the centroid of each "cell", and then get the cells from the centroids for each row.
![image](https://user-images.githubusercontent.com/4607414/115692974-ab04d600-a389-11eb-961f-494f003ba16f.png)

### Update 2:
I've encountered this case: 
![image](https://user-images.githubusercontent.com/4607414/115720099-ed88db80-a3a6-11eb-9156-f58c4dee0bb7.png)
Expected output is:
```
1-1, 2-1
None, 2-2
None, 2-3
```
`None` means not a cell there, and it makes the number of elements is even between rows which is convenient for further processing.
Note: the PR is updated

jsvine:
Hi @tungph, thanks for your interest in this library and thank you for the PR. I agree that `pdfplumber`'s current approach to complex tables is not optimal, but it is the result of not wanting to over-impose assumptions about any given table. The handling of merged cells in tables is a tricky topic, and one with a lot of edge cases — what may be a good solution for one set of tables may be a poor one for others. So I'll need to consider this suggestion and PR somewhat carefully, and think about (and test) other PDFs.

Relatedly: I have an idea for providing a richer representation of complex tables — one that would make the structure of the tables much clearer — but it would be a substantial change and so it will probably have to wait for `v0.6.0`.

tungph:
Thank you for the consideration, @jsvine. Please let me know if you found bug with my PR. I'm happy to provide fix.

Safrone:
I would love to have this as an option when extracting tables even if it is not the default behavior.

Issue #387, opened 2021-03-19
Add guidelines for submitting pull requests


jsvine:
At the very least, should indicate:

- That PRs should be submitted against `develop`, not `stable`
- That PRs should contain tests
- How to write new tests
- How to run tests
- How to run linter and reformat code

Anything else I'm forgetting?

alexreg:
This looks good to me as a new contributor.

samkit-jain:
Suggestions:
* Adding oneself to the [list of contributors](https://github.com/jsvine/pdfplumber#acknowledgments--contributors)
* Update [CHANGELOG.md](https://github.com/jsvine/pdfplumber/blob/stable/CHANGELOG.md)

jsvine:
Belatedly: Closed via #428.

PR #422, opened 2021-04-22
Duplicate value for merged cell instead of `None`

tungph:
See #420


tungph:
I saw #387:
- [x] That PRs should be submitted against develop, not stable
- [x] That PRs should contain tests
- [x] Run linter and reformat code
- [x] Adding oneself to the list of contributors
- [x] Update CHANGELOG.md

jsvine:
Hi @tungph, thanks for your interest in this library and thank you for the PR. I agree that `pdfplumber`'s current approach to complex tables is not optimal, but it is the result of not wanting to over-impose assumptions about any given table. The handling of merged cells in tables is a tricky topic, and one with a lot of edge cases — what may be a good solution for one set of tables may be a poor one for others. So I'll need to consider this suggestion and PR somewhat carefully, and think about (and test) other PDFs.

Relatedly: I have an idea for providing a richer representation of complex tables — one that would make the structure of the tables much clearer — but it would be a substantial change and so it will probably have to wait for `v0.6.0`.

tungph:
Thank you for the consideration, @jsvine. Please let me know if you found bug with my PR. I'm happy to provide fix.

tungph:
Stalled. Let's close this PR

PR #462, opened 2021-07-02
Explicitly typecast `fontname` and `text` fields to str for char objects

### LINKED ISSUE #461: i  find a bug when use dedupe_chars()
## Describe the bug

*A clear and concise description of what the bug is.*
this is my code:
#!/usr/bin/env python3
import pdfplumber
import base64
import re
from io import BytesIO

class PDF_Extract(object):
    def __init__(self, base64_data):
        self.base64_pdf = base64_data
        self.pdf_bytes = self.load()

    def load(self):
        content = re.sub(r'^data[^,]*?base64,', '', self.base64_pdf)
        #import ipdb;ipdb.set_trace()
        data = base64.b64decode(content)
        bytes_file = BytesIO()
        bytes_file.write(data)

        return bytes_file

    def words_extract(self):
        pdf_data = []
        pdf_obj = pdfplumber.open(self.pdf_bytes)
        for page in pdf_obj.pages:
            page_data = []
            import ipdb;ipdb.set_trace()
            words_list = page.dedupe_chars().extract_words()
            #https://github.com/jsvine/pdfplumber/issues/71
            for node in words_list:    
                item = [
                    float(node['x0']),
                    float(node['top']),
                    float(node['x1'] - node['x0']),
                    float(node['bottom'] - node['top']), node['text']
                ]
                page_data.append(item)
            pdf_data.append(page_data)
            page.close()
        return pdf_data

when i use pdf in attach，it trigger an error "TypeError: '<' not supported between instances of 'bytes' and 'str'"  at code "words_list = page.dedupe_chars().extract_words()"

please help me fix it
thank you very much

*Paste it here, or attach a Python file.*

## PDF file

*Please attach any PDFs [necessary]([患者B.pdf](https://github.com/jsvine/pdfplumber/files/6754122/B.pdf)) to reproduce the problem.*
[
[患者B.pdf](https://github.com/jsvine/pdfplumber/files/6754126/B.pdf)
](url)
*If you need to redact text in a sensitive PDF, you can run it through [JoshData/pdf-redactor](https://github.com/JoshData/pdf-redactor).*

## Expected behavior

## Actual behavior

*What actually happened, instead?*

## Screenshots

9D90A4B9-1698-4103-B2B1-7EC84F951242

## Environment

- pdfplumber version: [e.g., 0.5.22]    0.5.28 
- Python version: [e.g., 3.8.1] python3
- OS: [e.g., Mac, Linux, etc.]  linux

## Additional context

samkit-jain:
This PR fixes #461. The issue was happening because, in the PDF, there was a char object as
```python
{
    "fontname": b"\xcb\xce\xcc\xe5",
    "adv": Decimal("5.000"),
    "upright": True,
    "x0": Decimal("-146.900"),
    "y0": Decimal("782.030"),
    "x1": Decimal("-141.900"),
    "y1": Decimal("792.030"),
    "width": Decimal("5.000"),
    "height": Decimal("10.000"),
    "size": Decimal("10.000"),
    "object_type": "char",
    "page_number": 1,
    "stroking_color": [1, 1, 1],
    "non_stroking_color": [0, 0, 0],
    "text": "B",
    "top": Decimal("49.970"),
    "bottom": Decimal("59.970"),
    "doctop": Decimal("49.970"),
}
```
and since the `fontname` was of type `bytes` and not `str`, the `sorted` method threw an error as not all the `fontname`s were having the same type.

As a fix, I made it such that when fetching the `chars`, we do an explicit type cast to a string of fields `text` and `fontname`.

@jsvine My concern with this is that `pdfplumber` will be altering the `fontname` as provided in the PDF in which case, a better solution might be to only to do the typecast when calling the `sorted(...)` method in `yield_unique_chars(...)`. What do you think?


jsvine:
Thanks, @samkit-jain! A very interesting issue here. I've spent some time looking at it, and what follows is my understanding.

As you note, it's just one character that's causing problems, and the problem is due to the `fontname` property being represented as bytes rather than a string. Here's why I think that's happening:

- In the raw PDF, the `FontDescriptor` is defined as such: 

```
<</Type/FontDescriptor/FontName/BABGUC+#cb#ce#cc#e5/FontBBox[-7 -140 1000 859]/Flags 65568
/Ascent 859
/CapHeight 859
/Descent -140
/ItalicAngle 0
/StemV 150
/CIDSet 65 0 R
/FontFile2 53 0 R>>
```

- ... so the `/FontName` entry in that descriptor dictionary is the PostScript literal `/BABGUC+#cb#ce#cc#e5`

- When `pdfminer.six` (and `pdfminer` before it) encounters a `#` inside a PostScript literal, [it interprets the two subsequent characters as a hexadecimal-encoded byte](https://github.com/pdfminer/pdfminer.six/blob/22f90521b823ac5a22785d1439a64c7bdf2c2c6d/pdfminer/psparser.py#L348-L356). From what I understand, it also tries interpreting all PostScript literals as `utf-8` strings, but allows un-decodable strings to pass through as `bytes` objects. So I believe that's why `/BABGUC+#cb#ce#cc#e5` is coming through as `b"\xcb\xce\xcc\xe5"`. (Maybe there's a logic flaw that leads to the `BABGUC+` part being missing?)

From what I can tell online, however, PostScript `FontName`s are supposed to be strings and not contain `#-hex-hex` encodings, though I'm not 100% sure.

Since this is just one character in one PDF, and we haven't encountered this issue before (despite many people using `pdfplumber` to parse documents with non-Latin alphabets), I'd lean toward not adding code to handle it. Instead, the original issue might be addressed by adding these two lines inside the `for page` loop:

```python
for char in page.objects["char"]:
    char["fontname"] = str(char["fontname"])
```

What do you think?

samkit-jain:
Hi @jsvine This is very insightful. Yes, it could be that the font name was in a non-UTF-8 format, say UTF-16. I also tried repairing the PDF using Ghostscript but it didn't have any effect. I agree that it is a rare case but I am not sure if not adding a fix in the library is the right way to go. But, yes, can wait and see if more people report a similar issue.

jsvine:
Sounds good — let's wait to see whether this may be a more widespread issue with some PDFs.

ayhamkan:
> Thanks, @samkit-jain! A very interesting issue here. I've spent some time looking at it, and what follows is my understanding.
> 
> As you note, it's just one character that's causing problems, and the problem is due to the `fontname` property being represented as bytes rather than a string. Here's why I think that's happening:
> 
> * In the raw PDF, the `FontDescriptor` is defined as such:
> 
> ```
> <</Type/FontDescriptor/FontName/BABGUC+#cb#ce#cc#e5/FontBBox[-7 -140 1000 859]/Flags 65568
> /Ascent 859
> /CapHeight 859
> /Descent -140
> /ItalicAngle 0
> /StemV 150
> /CIDSet 65 0 R
> /FontFile2 53 0 R>>
> ```
> 
> * ... so the `/FontName` entry in that descriptor dictionary is the PostScript literal `/BABGUC+#cb#ce#cc#e5`
> * When `pdfminer.six` (and `pdfminer` before it) encounters a `#` inside a PostScript literal, [it interprets the two subsequent characters as a hexadecimal-encoded byte](https://github.com/pdfminer/pdfminer.six/blob/22f90521b823ac5a22785d1439a64c7bdf2c2c6d/pdfminer/psparser.py#L348-L356). From what I understand, it also tries interpreting all PostScript literals as `utf-8` strings, but allows un-decodable strings to pass through as `bytes` objects. So I believe that's why `/BABGUC+#cb#ce#cc#e5` is coming through as `b"\xcb\xce\xcc\xe5"`. (Maybe there's a logic flaw that leads to the `BABGUC+` part being missing?)
> 
> From what I can tell online, however, PostScript `FontName`s are supposed to be strings and not contain `#-hex-hex` encodings, though I'm not 100% sure.
> 
> Since this is just one character in one PDF, and we haven't encountered this issue before (despite many people using `pdfplumber` to parse documents with non-Latin alphabets), I'd lean toward not adding code to handle it. Instead, the original issue might be addressed by adding these two lines inside the `for page` loop:
> 
> ```python
> for char in page.objects["char"]:
>     char["fontname"] = str(char["fontname"])
> ```
> 
> What do you think?

still doesn't work all the time

jsvine:
Ended up implementing a similar, but slightly different version of this, based on some research about the common bytes-typed fontnames we saw cropping up in issues/discussions: https://github.com/jsvine/pdfplumber/pull/862/commits/9441ff7628fff9f69d81c6afd8ef439bf101b254

Thank you for getting the ball rolling on this, @samkit-jain!

PR #497, opened 2021-09-02
Fix slowdown in extract_words on long words (#483)

jsvine:
This commit/PR fixes the slowdown in `.extract_words(...)` and `WordExtractor.iter_chars_to_words(...)` which occurred for very long "words", and which was caused by repeatedly re-calculating bounding box. See #483 for discussion and example.

It also adds `utils.merge_bboxes(bboxes)`, which returns the smallest bounding box that contains all bounding boxes in the `bboxes` argument, and which we use in the fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,14 @@
 All notable changes to this project will be documented in this file. The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
 ## [0.5.29] - [unreleased]
+### Added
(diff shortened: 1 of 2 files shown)


samkit-jain (approved):

samkit-jain:
@jsvine Do we also have any time comparison stats to validate that it is now indeed faster? As per https://stackoverflow.com/a/19527487/7760998, we can also look at adding a test for it based on the PDF that was shared in the issue/discussion.

jsvine:
@samkit-jain Apologies for missing your comment here. I conducted some informal tests noted in https://github.com/jsvine/pdfplumber/discussions/483#discussioncomment-1269476 but nothing very rigorous. Certainly open to adding a test!

PR #504, opened 2021-09-16
fix chars deduplication for words with intentionally duplicated chars

konradmalik:
There is a problem with the current deduplication algorithm - it removes "intentionally" duplicated letters.

For example, if we run deduplication on chars `ssttiillll`, we will have `stil`, which is bad. We want `still`.

The solution is simple - instead of returning the first letter of each cluster of grouped characters, return every second character. This way, when the cluster has len=2 we will have 1 letter. When the cluster has len=4 we will have 2 letters. With clusters of len=1 we have 1 letter. Clusters of len=3 we have 2 letters which is neither good nor bad. In theory, this should not happen when the algorithm is run on a duplicated page.

This PR implements this change.

### DIFF
--- a/pdfplumber/utils.py
+++ b/pdfplumber/utils.py
@@ -200,7 +200,9 @@ def dedupe_chars(chars, tolerance=1):
         for grp, grp_chars in itertools.groupby(sorted_chars, key=key):
             for y_cluster in cluster_objects(grp_chars, "doctop", t):
                 for x_cluster in cluster_objects(y_cluster, "x0", t):
-                    yield sorted(x_cluster, key=pos_key)[0]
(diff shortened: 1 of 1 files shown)


samkit-jain:
Hi @konradmalik Appreciate you opening a pull request. Request you to update the PR as per the contribution guidelines mentioned at https://github.com/jsvine/pdfplumber/blob/stable/CONTRIBUTING.md#submitting-pull-requests

konradmalik:
@samkit-jain thank you for the info. I've made 3 mistakes:
- haven't read the contributing guidelines
- wrongly assumed the error was in that place when in reality it was an actual problem with my pdf which proper tests showed me
- created this PR too fast ;)
I'm closing this one

PR #515, opened 2021-10-15
Upgrade pdfminer.six from 20200517 to 20211012

jsvine:
See pdfminer.six's changelog for details: https://github.com/pdfminer/pdfminer.six/blob/develop/CHANGELOG.md

... but a key difference is an improvement in how it assigns `line`, `rect`, and `curve` objects. (Diagonal two-point lines, for instance, are now `line` objects instead of `curve` objects.)

As a result, this commit also adjusts some of the tests, where the pre-20211012 versions had been incorrectly assigning lines as `LTCurve` objects.

Note: This commit also tweaks CHANGELOG.md to call the next release `0.6.0`, since this will be a breaking change for some use-cases — though an improvement, and necessary if we don't want `pdfplumber` to fall far out of sync with `pdfminer.six`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,12 @@
 
 All notable changes to this project will be documented in this file. The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
-## [0.5.29] - [unreleased]
+## [0.6.0] - [unreleased]
 ### Added
 - Add `utils.merge_bboxes(bboxes)`, which returns the smallest bounding box that contains all bounding boxes in the `bboxes` argument.
 
 ## Changed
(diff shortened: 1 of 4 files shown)


samkit-jain (commented):
Long-awaited upgrade. @jsvine Should all the references to the `curves` property be removed as well if with this uprade, all curves would be lines? For example, https://github.com/jsvine/pdfplumber/blob/1328a00e13bb27b0bd8b8913f522b90af6fb2138/pdfplumber/container.py#L39

jsvine:
Good question, and I can see how that'd be confusing. Curve objects do still exist — they're paths with more than 2 points and which do not constitute a rectangle (or multiple consecutive rectangles). The fixes in the new version of `pdfminer.six` simply assign "LTLine" correctly in instances where it had previously been assigning "LTCurve," for instance diagonal lines. For an example of some paths that are still correctly curve objects, see the shapes on the final page of this PDF: http://www.pdfill.com/example/pdf_drawing_new.pdf

samkit-jain (approved):

Issue #463, opened 2021-07-05
Cannot decode contents of annotations


tungph:
## Describe the bug
While trying to get contents from an annotation, I got this error:
`UnicodeDecodeError: 'utf-8' codec can't decode byte 0xfe in position 0: invalid start byte`

The content for the first annotation is in Japanese. 

## Code to reproduce the problem

```python
    with pdfplumber.open('test.pdf') as pdf:
        print(pdf.annots)
```

## PDF file

[test.pdf](https://github.com/jsvine/pdfplumber/files/6762836/test.pdf)

## Expected behavior
```
{'uri': None, 'title': 'tung.phanhuy', 'contents': '日本語'}
{'uri': None, 'title': None, 'contents': None}
{'uri': None, 'title': 'tung.phanhuy', 'contents': '"well"'}
{'uri': None, 'title': None, 'contents': None}
{'uri': None, 'title': 'tung.phanhuy', 'contents': 'table'}
{'uri': None, 'title': None, 'contents': None}
```

## Environment

- pdfplumber version:  0.5.28
- Python version: 3.9.5
- OS: Mac

## Additional context
The content for the annotations is in both English and Japanese

tungph:
@jsvine @samkit-jain
To bypass the problem, I try to decode with `utf-16` if `utf-8` fail:
```python
    # page.py@125

    for k, v in extras.items():
        if v is not None:
            try:
                extras[k] = v.decode('utf-8')
            except UnicodeDecodeError:
                extras[k] = v.decode('utf-16')
```

jsvine:
Thank you for flagging this @tungph. I will look into this.

devWhyqueue:
Got the same problem trying to extract hyperlinks.

The following code:
```python
def extract_urls(self):
    with pdfplumber.open(self._file) as pdf:
        return {uri_obj['uri'] for uri_obj in pdf.hyperlinks}
```
raises an UnicodeDecodeError with this traceback:
```
 File "...", line 184, in extract_urls
    return {uri_obj['uri'] for uri_obj in pdf.hyperlinks}
  File ".../pdf.py", line 98, in hyperlinks
    return list(itertools.chain(*gen))
  File ".../pdf.py", line 97, in <genexpr>
    gen = (p.hyperlinks for p in self.pages)
  File ".../page.py", line 155, in hyperlinks
    return [a for a in self.annots if a["uri"] is not None]
  File ".../page.py", line 151, in annots
    return list(map(parse, raw))
[...]

PR #519, opened 2021-10-20
Handle utf-16-encoded annotations (#463)

jsvine:
Thanks to @tungph for the fix proposal.


samkit-jain (approved):
Approving the PR but this feels more like a hack. There could be a possibility in the future that we might have to add more encodings as defined [here](https://docs.python.org/3/library/codecs.html#standard-encodings). A more reliable solution would be a logic that automatically identifies the codec to be used (which I don't think is even possible to do with 100% accuracy). Or perhaps, adding a parameter for the users to pass in the encoding.

jsvine:
Thanks, and I agree: Not the world's most robust fix, but workable for now. I think the parameterizing is an intriguing idea, though I wonder: Are PDFs guaranteed to maintain encoding consistency throughout their annotations (or other content)? I suppose even if they do not, a user with a particularly tricky PDF could then try whatever combination of encodings suited them.

PR #520, opened 2021-10-20
Remove decimalizing (but let CLI adjust precision)

jsvine:
This PR updates the slightly outdated work on the [`un-decimalize` branch](https://github.com/jsvine/pdfplumber/tree/un-decimalize), making it compatible with the `develop` branch's changes since then. The main commit message from that branch:

```
Per discussion at https://github.com/jsvine/pdfplumber/discussions/346
and input from @ramcdona, this commit changes pdfplumber's behavior
regarding floating point numbers. Specifically, it removes all
conversion of floats to Decimal objects. This brings several advantages:

- Increased precision (where applicable)
- Decreased code complexity
- Increased performance (~10% speedup on test suite)
- Increased fidelity to `pdfminer.six` output

These seem to outweigh the disadvantages:

- Some tests break (but have been easily fixed) due to increased
  precision and/or floating point arithmetic artifacts
- Some users' scripts may also break, if they depend on strict equality
  testing, though these *should* also be easily fixable

Because some form of automatic rounding may still be desirable for the
pdfplumber CLI utility, the conversion methods (.to_csv, .to_json) have
been adjusted to accept a `precision` argument.
```

Tests are working with only minor revision, but edge-cases may warrant further scrutiny, as this will be a breaking change for some use-cases — especially those that try testing equivalence of two attributes via `==`.

### DIFF
--- a/pdfplumber/cli.py
+++ b/pdfplumber/cli.py
@@ -28,6 +28,8 @@ def parse_args(args_raw):
 
     parser.add_argument("--laparams", type=json.loads)
 
+    parser.add_argument("--precision", type=int)
+
     parser.add_argument("--pages", nargs="+", type=parse_page_spec)
 
     parser.add_argument(
@@ -43,7 +45,11 @@ def parse_args(args_raw):
 def main(args_raw=sys.argv[1:]):
     args = parse_args(args_raw)
     converter = {"csv": convert.to_csv, "json": convert.to_json}[args.format]
-    kwargs = {"csv": {}, "json": {"indent": args.indent}}[args.format]
+    kwargs = {
+        "csv": {"precision": args.precision},
+        "json": {"precision": args.precision, "indent": args.indent},
+    }[args.format]
+
     with PDF.open(args.infile, pages=args.pages, laparams=args.laparams) as pdf:
         converter(pdf, sys.stdout, args.types, **kwargs)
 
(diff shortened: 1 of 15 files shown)


pdfplumber/cli.py
```diff
@@ -28,6 +28,8 @@ def parse_args(args_raw):
 
     parser.add_argument("--laparams", type=json.loads)
 
+    parser.add_argument("--precision", type=int)
```
samkit-jain, line 31: The check should be for positive integers only. With just `int`, negative numbers will be accepted as well. Reference for implementation: https://stackoverflow.com/a/14117511/7760998

samkit-jain (changes requested):

pdfplumber/cli.py
jsvine: Hmmm, interesting! It may counterintuitive and hard to imagine a great use-case, but negative values seem, at least to me, that they could be legitimate for `precision`, since negative rounding is possible in Python. I.e.,:

```python
for i in reversed(range(-2, 3)):
    print(f"{i: >2}: {round(777.777, i):.3f}")
```

```
 2: 777.780
 1: 777.800
 0: 778.000
-1: 780.000
-2: 800.000
```

My instinct is to allow negative values. What do you think?

samkit-jain: Oh, I didn't know you could set negative values in the round operation. Makes sense then. Thanks.

samkit-jain (approved):

Issue #10, opened 2016-03-15
Is there a way to get something similar to pdftotext's layout?


dannguyen:
Is there an option similar to pdftotext's `-layout` flag, which "maintain[s] original physical layout"?  I understand that's probably up to the pdfminer engine...

Here's what I mean:
### Original PDF

http://www.supremecourt.gov/oral_arguments/argument_transcripts/07-1315.pdf

![image](https://cloud.githubusercontent.com/assets/121520/13770068/b6db3480-ea41-11e5-9bf3-bf389f9a5e8a.png)
### pdftotext with `-layout`

``` sh
$ curl -O http://www.supremecourt.gov/oral_arguments/argument_transcripts/07-1315.pdf
$ pdftotext -f 1 -l 1 -layout 07-1315.pdf -
```

Output:

```
                        Official - Subject to Final Review


 1      IN THE SUPREME COURT OF THE UNITED STATES

 2   - - - - - - - - - - - - - - - - - x

 3   MICHAEL A. KNOWLES,                            :

 4   WARDEN,                                        :

 5              Petitioner                          :

 6         v.                                       :        No. 07-1315

 7   ALEXANDRE MIRZAYANCE.                          :

 8   - - - - - - - - - - - - - - - - - x

 9                              Washington, D.C.

10                              Tuesday, January 13, 2009

11

12                  The above-entitled matter came on for oral

13   argument before the Supreme Court of the United States

14   at 1:01 p.m.

15   APPEARANCES:

16   STEVEN E. MERCER, ESQ., Deputy Attorney General, Los

17     Angeles, Cal.; on behalf of the Petitioner.

18   CHARLES M. SEVILLA, ESQ., San Diego, Cal.; on behalf

19     of the Respondent.

20

21

22

23

24

25


                                        1

                           Alderson Reporting Company
```
### pdfminer

``` py

import pdfplumber
import requests
fname = "/tmp/whatev.pdf"
resp = requests.get("http://www.supremecourt.gov/oral_arguments/argument_transcripts/07-1315.pdf")
with open(fname, 'wb') as f:
      f.write(resp.content)
pdf = pdfplumber.open(fname)
print(pdf.pages[0].extract_text())
```

Output:

```
Official - Subject to Final Review 
1 IN THE SUPREME COURT OF THE UNITED STATES 
2 - - - - - - - - - - - - - - - - - x 
3 MICHAEL A. KNOWLES,               : 
4 WARDEN,                           :
5  Petitioner            :
[...]

PR #532, opened 2021-11-03
Add experimental .extract_text(layout=True)

jsvine:
See this method and docstring for details regarding the implementation: https://github.com/jsvine/pdfplumber/blob/d235d4bbc4ad0c1b290916cae67060badf75198d/pdfplumber/utils.py#L344-L381

The approach is similar to @jsfenfen's helpful suggestion in https://github.com/jsvine/pdfplumber/issues/10#issuecomment-197071287 (thanks!), but with a some tweaks for fidelity and parameterization, as well as integration with the rest of the library.

Notes:

- The "density" defaults, by the way, are somewhat arbitrary and simply reflect an intuition after experimenting on a few PDFs.
- This PR adds two tests (for basic and cropped pages), but I'm open to adding more.

Other changes:

- `.extract_text(chars, ...)` returns "" if passed no characters, for return-type consistency.
- `.extract_words(...)` now includes `doctop` in its returned attributes.


samkit-jain (approved):

jsvine:
Thanks for the approval!

PR #545, opened 2021-11-22
Fix bug that crashed table extraction when null value provided for `(text|intersection)_(x|y)_tolerance` keys

samkit-jain:
Resolves bug raised in #539. The issue was happening because, in `TableFinder`, the settings were being cleaned up before the table extraction but in `extract_table(s)`, the updated table settings were not being used and passing in None in the tolerance setting resulted in a comparison between numeric type and None type and that caused the exception. The issue was resolved by creating a static method to cleanup the user-provided table settings and using it for every operation post table extraction.


jsvine:
Thanks, @samkit-jain! I think it does make sense to do some settings cleanup/resolution, and so I'm definitely leaning toward merging this PR. But I wonder: Is it always safe to assume that a user passing `None` for a setting value *means* they want to use the default value? And if they want to use the default value, why pass the setting at all?

samkit-jain:
I don't see any value in the settings that can have a meaningful null value and hence, think that the property not being provided or having the value as null can be treated the same. At https://github.com/jsvine/pdfplumber#table-extraction-settings, null value is being provided and that similar to in discussion #539 can confuse the users. We can also just remove the null and give those properties the default values.

jsvine (approved):

jsvine:
Thanks, @samkit-jain! That makes sense to me. And I agree: It's worth tweaking the README so that those `None` values in the table-settings documentation are listed as the default values. I can/will do that.

Issue #475, opened 2021-07-19
snap_x_tolerance and snap_y_tolerance for extra flexibility in table extraction


ratulbhadury:
Hi Jeremy

Before I get into my issue, I just want to say (again), thanks for all your efforts on this package. I've found it much easier to use than many of the other similar python libraries.

I'm using v0.5.28. I'm faced with a scenario where I would like to specify different tolerances for the x and y directions when it comes to 'snapping' parallel lines into one. 

I see the [exact same request](https://github.com/jsvine/pdfplumber/issues/52) had been made some years ago, and it looks as it had been accepted and implemented back then into the v0.6 branch, but hasn't made its way into the master one.

Do you have plans to include this into the master branch? If not, can you suggest any workarounds for how I could get this to work?

Thanks and kind regards,

Ratul

samkit-jain:
Yes, it is a useful feature request. Feel free to raise a PR if you have a solution in mind.

ratulbhadury:
Hi @samkit-jain 

I haven't investigated a solution yet, but I have seen that this feature has been previously requested, and is in fact part of the v0.6 alpha branch.

I have not yet looked deeply into the architecture of the v0.6 branch and its differences with the stable branch, but would it be possible to include this feature that is already in that branch, and merge it into the stable branch? Please refer to PR #51. 

Many thanks

jsvine:
Hi @ratulbhadury, and thanks again for this suggestion. Your requested feature is now available in `v0.6.0`! See https://github.com/jsvine/pdfplumber/pull/553 for details.

PR #51, opened 2018-02-20
Add snap_x_tolerance and snap_y_tolerance

dustindall:
Hi @jsvine ,

I was working with a table where i needed to additional snap_tolerance flexibility on the y-axis relative to the x-axis. Therefore, I added a snap_x_tolerance and snap_y_tolerance parameters in the same fashion as text_tolerance and intersection_tolerance.

Thank you for working on this. It is a great package.

Dustin

PR #553, opened 2021-12-02
Add snap_x/y_tolerance and join_x/y_tolerance table-extraction settings

jsvine:
Based largely on @dustindall's work in PR #51, adapted to current code. Also resolves issue #475.

This PR also changes `table.merge_edges(...)` behavior when `join_tolerance` (and `x`/`y` variants) `<= 0`, so that joining is attempted regardless, to handle cases of overlapping lines (which would still fulfill `tolerance == 0`). This is an uncommon situation but is, I believe, represents the user-expected behavior.

### DIFF
--- a/pdfplumber/table.py
+++ b/pdfplumber/table.py
@@ -8,16 +8,20 @@ DEFAULT_MIN_WORDS_VERTICAL = 3
 DEFAULT_MIN_WORDS_HORIZONTAL = 1
 
 
-def snap_edges(edges, tolerance=DEFAULT_SNAP_TOLERANCE):
+def snap_edges(
+    edges, x_tolerance=DEFAULT_SNAP_TOLERANCE, y_tolerance=DEFAULT_SNAP_TOLERANCE
+):
     """
     Given a list of edges, snap any within `tolerance` pixels of one another
     to their positional average.
     """
-    v, h = [list(filter(lambda x: x["orientation"] == o, edges)) for o in ("v", "h")]
+    by_orientation = {"v": [], "h": []}
+    for e in edges:
+        by_orientation[e["orientation"]].append(e)
 
-    snap = utils.snap_objects
-    snapped = snap(v, "x0", tolerance) + snap(h, "top", tolerance)
-    return snapped
+    snapped_v = utils.snap_objects(by_orientation["v"], "x0", x_tolerance)
+    snapped_h = utils.snap_objects(by_orientation["h"], "top", y_tolerance)
+    return snapped_v + snapped_h
 
 
 def join_edge_group(edges, orientation, tolerance=DEFAULT_JOIN_TOLERANCE):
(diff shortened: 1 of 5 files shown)


pdfplumber/table.py
```diff
@@ -59,16 +65,18 @@ def get_group(edge):
         else:
             return ("v", edge["x0"])
 
-    if snap_tolerance > 0:
-        edges = snap_edges(edges, snap_tolerance)
+    if snap_x_tolerance > 0 or snap_y_tolerance > 0:
```
samkit-jain, line 68: Shouldn't it be an `and` comparator? And should <=0 values be treated differently?

pdfplumber/table.py
```diff
@@ -59,16 +65,18 @@ def get_group(edge):
...
+        edges = snap_edges(edges, snap_x_tolerance, snap_y_tolerance)
 
-    if join_tolerance > 0:
-        _sorted = sorted(edges, key=get_group)
-        edge_groups = itertools.groupby(_sorted, key=get_group)
-        edge_gen = (
-            join_edge_group(items, k[0], join_tolerance) for k, items in edge_groups
+    _sorted = sorted(edges, key=get_group)
```
samkit-jain, line 71: Should the `join_x_tolerance > 0` and `join_y_tolerance > 0` check be added here as well?

pdfplumber/table.py
jsvine: I don't think so? I think you want to run this block as long as *any* positive snap tolerance is specified. Or perhaps I misunderstand the concern?

Re. `<=0` values:

- `0` should be treated, I think, as indicating "do not snap" and thus no additional code should need to be run.

- I suppose we should convert negative values to `0`. I'll push some code that does that. Unless you see a strong purpose for a negative tolerance here?

pdfplumber/table.py
jsvine: My apologies, I could/should have been clearer about this. This is what I meant by this part of the PR message:

> This PR also changes `table.merge_edges(...)` behavior when `join_tolerance` (and `x`/`y` variants) `<= 0`, so that joining is attempted regardless, to handle cases of overlapping lines (which would still fulfill `tolerance == 0`). This is an uncommon situation but is, I believe, represents the user-expected behavior.

Essentially, `join_edge_group(...)` is still helpful, even when `join_tolerance==0`, because it combines ("joins") overlapping line segments that lay along the same infinite line.

pdfplumber/table.py
jsvine: On negative values, decided instead just to raise an error if they're negative — which gives us more flexibility to change the behavior in the future (rather than silently modifying the user's initial parameters). See https://github.com/jsvine/pdfplumber/pull/553/commits/aa2d594d3b3352dbcef503e4df2e045d69fc2511 below.

jsvine:
Merged, but I'm open to reverting/changing anything that seems off before releasing the next version.

PR #568, opened 2021-12-17
Make final changes for v0.6.0

jsvine:
See commits below. In short:

- Updates dependencies
- Adds `isort` linting/formatting
- Removes deprecated `.load(...)` method


jsvine:
Note: Code coverage decreases ever so slightly because we have removed a deprecated method. I think that's fine.

samkit-jain (approved):

PR #581, opened 2022-01-13
addressing issue #578

### LINKED ISSUE #578: Context manager doesn't close file on exceptions
## ```pdfplumber.open()``` doesn't seem to be releasing the file when an error occurs in intself

I am trying to make my code robust to exceptions and good at logging them.
I needed some files to test exceptions with, so i created a new empty text document and renamed it to "empty.pdf".

My code's workflow is:
- scan a directory full of PDFs
- for each PDF:
    - ```pdfplember.open()``` it
    - extract text, tables, and whatever else
    - move the file to an "old" directory with ```os.rename```

to keep it simple I cut the input and output folders from the equations, as well as the loop on the input folder and the logging part

```
import pdfplumber
import os

try:
    with pdfplumber.open("empty.pdf") as pdf:
        pdf.do_something() # it won't even reach this point, you are good to copy and run the whole code
except Exception:
    # log the exception and keep the process going
    pass
os.rename("empty.pdf", "_empty.pdf")
```

## PDF file

Again:
just create an empty text document and rename it to "empty.pdf"

## Expected behavior

The file should be moved to the new location (in this simplified version, just renamed), even if an exception occurred

## Actual behavior

The program throws a ```PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'empty.pdf' -> '_empty.pdf'```
**BUT** if you open a "good" pdf, and then raise an exception inside the with-block, the program works as intended.

## Environment

- pdfplumber version: 0.6.0
- Python version: 3.10
- OS: Windows 10

## Additional context

If in the ```except:``` block I open another file (wheter an empty PDF or not), the ```rename``` part works flawlessly, and this is making me think that when you call ```pdfplumber.open()``` it "resets" all handlers it had before.
Unfortunately I didn't try with a corrupted PDF because i don't have one at this time, but my goal was to catch exception raised because of bad PDFs. I suppose the problem would be present even if you were to use a currupted PDF

augeos-grosso:
This PR fixes #578 .
Indeed the line causing the missed close() call on Windows (and only on Windows it seems) was the one @samkit-jain pointed out.

The original exception was ```pdfminer.pdfparser.PDFSyntaxError```, but i figured it would be best using ```pdfminer.psparser.PSException```, since it is the most general exception i could find that made sense using.

### DIFF
--- a/README.md
+++ b/README.md
@@ -438,7 +438,7 @@ Many thanks to the following users who've contributed ideas, features, and fixes
 - [Daniel Peña](https://github.com/trifling)
 - [bobluda](https://github.com/bobluda)
 - [@ramcdona](https://github.com/ramcdona)
-
+- [augeos-grosso](https://github.com/augeos-grosso) aka [johnhuge](https://github.com/johnhuge)
 ## Contributing
 
 Pull requests are welcome, but please submit a proposal issue first, as the library is in active development.
--- a/README.md
+++ b/README.md
@@ -438,7 +438,7 @@ Many thanks to the following users who've contributed ideas, features, and fixes
 - [Daniel Peña](https://github.com/trifling)
 - [bobluda](https://github.com/bobluda)
 - [@ramcdona](https://github.com/ramcdona)
-- [augeos-grosso](https://github.com/augeos-grosso) aka [johnhuge](https://github.com/johnhuge)
+- [johnhuge](https://github.com/johnhuge)
 
 ## Contributing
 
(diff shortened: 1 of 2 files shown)


jsvine:
Thanks for filing this @augeos-grosso! Two quick things:

- Could you push an empty/minor commit? It looks like the PR-checking workflow got tripped up on a network issue. We can squash the commit when we merge.

- Thanks for adding yourself to the contributor's list — well deserved! But could you tweak it (i.e., choose just one link/name/handle) for the purposes of consistency with the other contributors in the list? Doing that would actually take care of the needs-another-commit-to-rerun-workflow situation above.

jsvine:
Thanks! And, hmmm, workflow still seems gummed up. I'll take a closer look later today; I could just run the tests locally if it comes to that.

augeos-grosso:
Of course!
I'm here to help if you need me

jsvine:
Thanks again @augeos-grosso! Ran the tests locally — they passed, but coverage decreased since there were no tests added for this change. That's fine. They're simple tests. I've now written them and will push them to `develop`.

PR #657, opened 2022-05-24
Add py.typed marker for PEP 561 compatibility

jhonatan-lopes:
Although type hints have been introduced on the previous release (thank you very much for that! :pray:), `mypy` does not detect the type hints when the package is installed since it is missing stubs or a `py.typed` marker:

```console
src/my_package/myfile.py:4: error: Skipping analyzing "pdfplumber": module is installed, but missing library stubs or py.typed marker
```
I propose adding an empty `py.typed` file, which should be enough for `mypy` to detect that `pdfplumber` is typed, and adding this file to the setup's `package_data`. This should make `pdfplumber` PEP 561 compatible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file. The format 
 - Add `pdfplumber.ctm` submodule with class `CTM`, to calculate scale, skew, and translation of a current transformation matrix obtained from a `char`'s `"matrix"` property.
 - Add `page.search(...)`, an *experimental feature* that allows you to search a page's text via regular expressions and non-regex strings, returning the text, any regex matches, the bounding box coordinates, and the char objects themselves. ([#201](https://github.com/jsvine/pdfplumber/issues/201))
 - Add `--include-attrs`/`--exclude-attrs` to CLI (and corresponding params to `.to_json(...)`, `.to_csv(...)`, and `Serializer`.
+- Add `py.typed` for PEP561 compatibility and detection of typing hints by mypy.
 
 ### Removed
 
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,2 +1,3 @@
 include LICENSE.txt
 include README.md
+include pdfplumber/py.typed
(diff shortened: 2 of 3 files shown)


jsvine:
Thanks @jhonatan-lopes, and good catch! Two quick questions:

- What is the purpose of including `typed.py` in `package_data`, given that `typed.py` is already in the main `pdfplumber` subdirectory?

- In any case, my understanding is that repositories should use `package_data` *or* `MANIFEST.in`, but not both. Given that this repository already has a `MANIFEST.in` file, what is the advantage of using `package_data` here?

jhonatan-lopes:
Hi @jsvine, thanks for this amazing library! 

Although `py.typed` is in the main directory, it must be included in as the package data to make sure that it is present when the package is installed. If this is not included, then once it is installed, only the source code will be copied over and `mypy` won't be able to detect that the package has type hints because it won't have a `py.typed` file or stubs (`py.typed` is not detected as source code since it is not a `.py` file).

Now, how this file is included is a matter of choice, I think. [PEP 561](https://peps.python.org/pep-0561/#packaging-type-information) has an example with the `package_data` formatting, so I ended up using that. As long as the file is included in the package installation, I don't think it matters. Both options are valid, but I don't think there should be any problems mixing the `package_data` with the existing `MANIFEST.in` (see https://stackoverflow.com/questions/24727709/do-python-projects-need-a-manifest-in-and-what-should-be-in-it and https://packaging.python.org/en/latest/guides/using-manifest-in/)

I can change the include to use the existing `MANIFEST.in` if you prefer.

jsvine:
Thanks, @jhonatan-lopes! That's a very helpful explanation. And, yes, lets use the existing MANIFEST.in instead. Feel free to force-push a squashed commit.

jhonatan-lopes:
Hi @jsvine, it is done! I'll let you deal with the merging of the PR instead of doing a force-push, if that's ok.

jsvine:
Super, thanks! Now merged, and will be adding a thanks to you in the README.md file.

Issue #683, opened 2022-07-06
Page.search ValueError: min() arg is an empty sequence


bpugnaire:
## Describe the bug

When using the page.search method with a regex, the program may raise a ValueError: min() arg is an empty sequence.

## Code to reproduce the problem

The issue is not easily reproducible, on some pages with the same Regex pattern I get the expected behavior (list of match or an empty list if no match). And sometimes I get the error

## Expected behavior

Get a list of match or an empty list if there are no matches

## Actual behavior

Got the ValueError: min() arg is an empty sequence

## Environment

- pdfplumber version:  0.7.1
- Python version: 3.8.1
- OS: Windows 10

## Additional context

Traceback (most recent call last):

  File "c:\Users\xxx\.virtualenvs\project\lib\site-packages\pdfplumber\page.py", line 316, in search
    return text_layout.search(pattern, regex=regex, case=case)
  File "c:\Users\xxx\.virtualenvs\project\lib\site-packages\pdfplumber\utils.py", line 528, in search
    return list(map(match_to_dict, gen))
  File "c:\Users\xxx\.virtualenvs\project\lib\site-packages\pdfplumber\utils.py", line 499, in match_to_dict
    x0, top, x1, bottom = objects_to_bbox(chars)
  File "c:\Users\xx\.virtualenvs\project\lib\site-packages\pdfplumber\utils.py", line 207, in objects_to_bbox
    min(map(itemgetter("x0"), objects)),
ValueError: min() arg is an empty sequence

bpugnaire:
To be noted, I only encountered the bug when using Regex = True

samkit-jain:
Hi @bpugnaire Appreciate your interest in the library and thanks for raising this issue. Could you please provide the sample code that you used to reproduce this issue? It would help us investigate and fix. If possible, the PDF that resulted in this behaviour as well.

bpugnaire:
match_group = '(Figure|Fig\.|fig\.|Tab\.|tab\.|Tabl\.)'
search_result = page.search("(?<!(\(|\[))"+match_group, regex=True)

The regex is here to capture "xxx Figure 1xxx" and not "xxx [Figure1 xxx"

jsvine:
Thanks @bpugnaire. Are you able to share the PDF? That would make it easier to diagnose the situation.

jsvine:
This is fixed in the above commit, and available as of v0.7.1.

PR #687, opened 2022-07-13
Handle `ValueError` exception when searching for text using regex

samkit-jain:
This PR attempts to fix the issue #683 reported by @bpugnaire. The issue was happening because 
```python
chars = [c for (text, c) in subset if c is not None]
```
would return an empty array because of which the subsequent
```python
x0, top, x1, bottom = objects_to_bbox(chars)
```
call would throw a `ValueError` exception.

I was able to reproduce the issue on the file [issue-71-duplicate-chars-2.pdf](https://github.com/jsvine/pdfplumber/blob/develop/tests/pdfs/issue-71-duplicate-chars-2.pdf) using the regex `\d+`. The solution implied is a simple one with `try-except` handling. However, I am unsure if this is the correct solution because it would result in the output having objects with `x0`, `x1`, ... being null which may cause further processing issues as they are expected to be numbers. Another solution would be to not include these objects in the final answer but that may result in incomplete data being returned. Yet to debug why this issue happened in the first place, that is, a solution that would result in the `chars` list not being empty.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to this project will be documented in this file. The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
+## Unreleased
+
+### Fixed
+
+- Fix `ValueError` exception when using `page.search(regex=True)`. ([#683](https://github.com/jsvine/pdfplumber/issues/683)
+
 ## [0.7.1] - 2022-05-31
 
 ### Fixed
--- a/pdfplumber/utils.py
+++ b/pdfplumber/utils.py
@@ -496,7 +496,10 @@ class TextLayout:
         def match_to_dict(m: Match[str]) -> Dict[str, Any]:
             subset = self.layout_tuples[m.start() : m.end()]
             chars = [c for (text, c) in subset if c is not None]
-            x0, top, x1, bottom = objects_to_bbox(chars)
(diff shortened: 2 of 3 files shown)


jsvine:
Many thanks for finding a PDF that allows us to reproduce the problem, and for writing the test, @samkit-jain! As you suspected, the `try-except` approach treated the symptoms of a deeper issue, rather than the cause. But your PR really helped to investigate the problem, which ended up being in `LayoutEngine.calculate(...)`, which had been assuming that `len(char["text"])` would always equal 1, which is not true for ligatures. (The test PDF contains the `ﬁ` ligature.) This created a mismatch between the string that `.search(...)` was using and the character-order table (`LayoutEngine.layout_tuples`). The fix turned out to be relatively simple, which is to iterate over all letters in `char["text"]`: https://github.com/jsvine/pdfplumber/commit/12feadb8fdbd86fb2f596abe803a1a46d58a58da

Worth noting: This fix uses the `same` char object data for both the `f` and `i` in the `fi` ligature (as an example), rather than, e.g., splitting recalculating the bounding box and other information as if the ligature had been split in half. Given the somewhat limited purpose of that step, I think that's probably OK, but a side effect will be that search-matches beginning or ending with just one letter in the ligature will have a bounding box that spans both.

Thanks again!

Issue #698, opened 2022-08-01
pip whl missing `py.typed`


jhonatan-lopes:
## The bug

#657 introduced a `py.typed` marker to indicate to mypy that the package is typed. However, the [built distribution on pypi](https://pypi.org/project/pdfplumber/0.7.4/#files) (wheel file) is missing the `py.typed` file, whereas the source distribution is including it, as it should.

When building the package locally with `pip wheel .` (with my fork of `pdfplumber`), the package builds successfully and `py.typed` is included in the build and the wheel file. 

## Expected behavior

The `py.typed` file maker should be included in the pypi wheel distribution.

## Actual behavior

The `py.typed` file maker is missing from the built distribution on pypi.

## Screenshots

First, the `.tar.gz` source distribution showing the `py.typed` file:

![image](https://user-images.githubusercontent.com/61277874/182208440-6315feee-a7df-4f50-a745-e2e8728e4f68.png)

Now, the `.whl` built distribution missing the `py.typed` file:

![image](https://user-images.githubusercontent.com/61277874/182208560-4e4c83bc-eeca-4aad-a616-7b316ff089a7.png)

The locally built `.whl` distribution:

![image](https://user-images.githubusercontent.com/61277874/182208723-f793faa9-4934-4970-a542-94456f4b4ed7.png)

## Environment

- pdfplumber version: 0.7.4
- Python version: 3.10
- OS: Manjaro Linux

jsvine:
Hi @jhonatan-lopes, and thanks for raising this issue. I'm not much of a packaging/distribution expert, so I have a couple of questions to help clarify my understanding:

- It makes sense to me that the *source* distribution would/should have `py.typed`.  But why would/should the built distribution have it? My understanding is that `py.typed` is primarily useful for working with the source files, but I may simply be misunderstanding.

- To build the distributions, I run `python -m build` (based on [these recommendations](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html)). Is that also what you're running? Or a different command?

jhonatan-lopes:
Hi @jsvine! I'm not a distributions expert in any way either :) but I'm trying my best to get typing annotations working for this amazing library, especially given the trouble with putting up the annotations in the code in first place. Packaging should have been the easy part.

To your questions:

[...]

PR #703, opened 2022-08-05
Mypy compatibility

jhonatan-lopes:
Add `py.typed` to package data to ensure that it is included when building. Include `zip_safe=False` per `mypy`'s instructions. See #698.


jhonatan-lopes:
@jsvine, could you please test this with your usual build process and check whether `py.typed` is now included in the source and built distributions?

jsvine:
> @jsvine, could you please test this with your usual build process and check whether `py.typed` is now included in the source and built distributions?

Yep, that fixed the issue! Many thanks, merging 👍

PR #732, opened 2022-09-19
Update README.md

tjex:
PyPDF2 is certainly still active. A statement that it seems to no longer be maintained is a bit of a cold shoulder (and in this case a false observation), and may well turn devs away from considering to use other peoples hard work.


jsvine:
Thanks for flagging @tjex. Glad to see that PyPDF2 is releasing new versions again. At the time I wrote that description (August 2020), there [hadn't been a release in more than four years](https://pypi.org/project/PyPDF2/#history) (May 2016). The statement was not intended as judgmental on anyone or their work (I happen to have a good number of inactive projects myself), but I understand how it could be perceived as such. I'll revise that entry.

tjex:
Not a problem! Thanks for the quick response and for pdfplumber!

PR #791, opened 2023-01-19
fix typing hints to include io.BytesIO

conitrade-as:


jsvine:
Thanks, looks good! Merging.

PR #811, opened 2023-02-13
fix: calculate rotation using 0 when Rotate is None

toshi1127:
An error occurred when calculating rotation when Rotate was None.
Fixed to calculate rotation using 0 when Rotate is None.

```
TypeError: unsupported operand type(s) for %: 'NoneType' and 'int'

self.rotation = _rotation % 360
File "/var/task/pdfplumber/page.py", line 54, in __init__
```

### DIFF
--- a/pdfplumber/page.py
+++ b/pdfplumber/page.py
@@ -90,7 +90,7 @@ class Page(Container):
         self.root_page = self
         self.page_obj = page_obj
         self.page_number = page_number
-        _rotation = resolve_all(self.page_obj.attrs.get("Rotate", 0))
+        _rotation = resolve_all(self.page_obj.attrs.get("Rotate", 0)) or 0
         self.rotation = _rotation % 360
         self.page_obj.rotate = self.rotation
         self.initial_doctop = initial_doctop


samkit-jain:
Hi @toshi1127 Thanks for raising the PR. Could you please add a test case with the PDF for the issue that you are facing which fails otherwise but passes with your fix?

toshi1127:
@samkit-jain 
Thanks for reviewing.

Is it mandatory to add the PDF as a test case to merge this PR?
We are developing SaaS, and we would like to avoid adding it as test data as much as possible, because an error occurred in a PDF uploaded by contractor.

If you insist, we will attempt to see if we can create a separate PDF to reproduce the error.

jsvine:
Thank you, @toshi1127  and thank you @samkit-jain. Ideally, PRs come with proving test cases, but this change is simple/clear enough, and true to the specification\*, that I think we can make an exception here. Merging.

\* Per the [spec](https://ghostscript.com/~robin/pdf_reference17.pdf) (p. 146), rotation refers to "The number of degrees by which the page should be rotated clockwise when displayed or printed. The value must be a multiple of 90. Default value: 0."

jsvine:
And I'm curious, @toshi1127: What's the SaaS product?

toshi1127:
@jsvine 
Thank you for the review !
I am using pdfplumber for the following product.

https://bakuraku.jp/denshichobo

PR #835, opened 2023-03-06
Correct placement of the `close()` method in the documentation

### LINKED ISSUE #834: Documentation of `close()`
Please describe, in as much detail as possible, your proposal and how it would improve your experience with pdfplumber.

The documentation (./README.md) suggests the `Page` class has a `close()` function.
It seems like only the `PDF` class has a `close()` function.

samkit-jain:
The method belongs to the `PDF` class and not the `Page` class and was in the incorrect section in the README file as pointed out by @sujayvadlakonda

Fixes #834


jsvine:
Thanks, good catch!

PR #856, opened 2023-04-07
update to the correct key name

Kalmaegi:
update to the correct key name and remove the unnecessary code


jsvine:
Thanks! Good catch and helpful README fix. (And do worry about the failing tests here; that's caused by a recent release of `Pillow`; I'll fix separately.)

Issue #909, opened 2023-06-20
Accessibility tagging


NathanTech7713:
Hi there,

Was wondering if, when the dev is particularly bored, would you mind considering implementing extraction of accessibility tagging?

Thank youPlease describe, in as much detail as possible, your proposal and how it would improve your experience with pdfplumber.

jsvine:
Hi @NathanTech7713 — thanks for your interest in this library, and for this suggestion. For my own notes and for others who may be less familiar:

- "Tagged PDFs" are documents that use the PDF spec's features for enabling accessibility, by adding semantic markup in the form of "tags": https://www.pdfa.org/wp-content/uploads/2019/06/TaggedPDFBestPracticeGuideSyntax.pdf

- `pdfminer.six`, the parser on which `pdfplumber` depends, does seem to have some functionality for identifying and extracting these tags — however, that functionality [comes from the `TagExtractor` class](https://github.com/pdfminer/pdfminer.six/blob/5114acdda61205009221ce4ebf2c68c144fc4ee5/pdfminer/pdfdevice.py#L242), which is a subclass of the `PDFDevice` class we use. Still, it's helpful to know that there's some functionality there, even if we'd have to patch it in.

And some general questions: What should the output of this extraction look like? A nested tree of tags? Something else? 

@NathanTech7713: Do you have any examples of other PDF extraction libraries that have a feature like this, and which you think would provide a useful model?

dhdaines:
Hi! I was about to make this same feature request.  I've done a bit of exploration here as I am working on extracting the structure from PDFs and, obviously, it makes sense to use *explicit* structure if it's there... well, sort of.

Most of the libraries that support tagged PDF are closed-source, but some functionality to extract it exists in [Poppler](https://poppler.freedesktop.org/) and [pdf.js](https://mozilla.github.io/pdf.js/),  and you can see the tags by running `pdfinfo -struct` on a PDF (or `pdfinfo -struct-text` to see the content of the tags as well).  Unfortunately the generation of structure and tags is, to put it mildly, highly variable across different PDF authoring tools, and I haven't come remotely close to understanding the (very convoluted) specification.  The W3C has a nice overview of logical structure and tagged PDF here: https://www.w3.org/TR/2014/NOTE-WCAG20-TECHS-20140408/pdf_notes.html

[...]

PR #937, opened 2023-07-19
Add support for structure tree and marked content sections

dhdaines:
Implements #909 with an IMHO rather convenient interface for marked content - IDs are listed in the structure tree, then propagated to objects in each page.


dhdaines:
Note that you can link the structure tree and text from marked content sections like this:
```python
def get_text_by_mcid(page):
    mcids = []
    for c in page.chars:
        mcid = c.get("mcid")
        if mcid is None:
            continue
        while len(mcids) <= mcid:
            mcids.append("")
        mcids[mcid] += c["text"]
    return mcids

def get_structure_tree_with_text(page):
    texts = get_text_by_mcid(page)
    st = page.structure_tree
    d = deque(st)
    while d:
        el = d.popleft()
        if "children" in el:
            d.extend(el["children"])
        if "mcids" in el:
            el["mcids"] = [texts[mcid] for mcid in el["mcids"] if mcid < len(texts)]
    return st
```
Not sure if this is helpful to have as a method in the `Page` class?  One thing I notice is that MCIDs are *not* reliably aligned to word breaks, they often change in the middle of a word for no apparent reason.

dhdaines:
Another helpful example, if you want to for instance get the bounding box of a `Table` element on a page:
```python
def get_tables(page):
    st = page.structure_tree
    d = deque(st)
    while d:
        el = d.popleft()
        if "children" in el:
            d.extend(el["children"])
        if el["type"] == "Table":
            yield el

def get_child_mcids(el):
    d = deque([el])
    while d:
        el = d.popleft()
        if "children" in el:
            d.extend(el["children"])
        if "mcids" in el:
            yield from el["mcids"]

t = next(get_tables(page))
mcids = set(get_child_mcids(t))
tbox = pdfplumber.utils.objects_to_bbox([
    c for c in itertools.chain(page.chars, page.images) if c.get("mcid") in mcids
])
```

dhdaines:
Another note - there is a small problem with this PR which is that there can be marked content sections which aren't referenced by the structure tree - this is specifically the case for headers and footers.  The PR retains their MCIDs but nothing else so there isn't any way to detect them.  I'll add a `marked_content_sections` property which contains this information.

dhdaines:
Another note - because this uses `pypdfium2` to get the structure tree it introduces quite a lot of overhead, though only if you choose to access it, since now you are reading (at least partially) each page twice, once with `pdfminer.six` and once with `pypdfium2`.  I will perhaps try to reimplement it with `pdfminer.six` - the logic of resolving the structure tree is slightly complicated, but the `pdf.js` implementation is a good guide.

pdfplumber/page.py
```diff
@@ -115,6 +119,42 @@ def normalize_color(
...
+        else:
+            self.cur_mcid = None
+
+    def end_tag(self) -> None:
+        self.cur_mcid = None
+
+    def tag_cur_item(self, item_type: Any) -> None:
+        # Implementation Inheritance Considered Harmful
+        cur_obj = self.cur_item._objs[-1]
+        assert isinstance(cur_obj, item_type)
```
jsvine: I'm curious: What is the bigger-picture purpose of this assert statement?

pdfplumber/page.py
```diff
@@ -115,6 +119,42 @@ def normalize_color(
...
+    def render_char(self, *args, **kwargs) -> float:  # type: ignore
+        adv = super().render_char(*args, **kwargs)
+        self.tag_cur_item(LTChar)
+        return adv
+
+    def render_image(self, *args, **kwargs) -> None:  # type: ignore
+        super().render_image(*args, **kwargs)
+        self.tag_cur_item(LTImage)
+
+    def end_figure(self, *args, **kwargs) -> None:  # type: ignore
```
jsvine: `pdfplumber` does not return `LTFigure` objects, which are somewhat confusingly named. Like `LTTextBoxHorizontal` (and similar), they refer not to discrete objects on the page, but rather layout-analyzed agglomerations of them. So I think we can skip handling these, but perhaps I misunderstand the intent here?

pdfplumber/page.py
```diff
@@ -115,6 +119,42 @@ def normalize_color(
...
+        assert isinstance(cur_obj, item_type)
+        # Hackery, which would not be necessary if pdfminer.six supported MCIDs
+        cur_obj.mcid = self.cur_mcid
+
+    def render_char(self, *args, **kwargs) -> float:  # type: ignore
+        adv = super().render_char(*args, **kwargs)
+        self.tag_cur_item(LTChar)
+        return adv
+
+    def render_image(self, *args, **kwargs) -> None:  # type: ignore
```
jsvine, line 155: What is the reason for tagging `image` and `char` objects, but not `line`, `rect`, or `curve` objects?

pdfplumber/page.py
dhdaines: Ah, it (and the associated `item_type` parameter) is just there to confirm my understanding of how `pdfminer.six` works internally - as the comment above notes this depends on the internal behaviour of the `render_*` methods.  In an ideal world they would simply return the objects they create instead of modifying some internal state.

pdfplumber/page.py
dhdaines: Ah, this could be an omission on my part (I didn't have a PDF with lines/rects/curves in a Figure handy to test with, but I should be able to create one myself with LibreOffice).  Thanks for catching it, I'll add a test case.

jsvine:
> Another note - because this uses pypdfium2 to get the structure tree it introduces quite a lot of overhead, though only if you choose to access it, since now you are reading (at least partially) each page twice, once with pdfminer.six and once with pypdfium2. 

Thanks for flagging. As I understand it, this PR introduces two separate-but-related features:

1. Adding the `mcid` attribute to each parsed object, which the PR currently handles entirely through subclassing `pdfminer.six`'s `PDFPageAggregator`. 
2. Extracting a PDF's structure tree, which depends on `pypdfium2`, which introduces the overhead you mentioned, and appears to fail on Python 3.8 (something related to `_ctypes.PyCArrayType` but I haven't investigated closely).

Would it be _possible_ to separate this PR into those two distinct features? Is that something you'd be _open to_?

pdfplumber/page.py
dhdaines: I'll have to get back to you on this one ... in the structure tree, the MCID is associated with the Figure, so the expected behaviour would be to have all the objects inside it be tagged with that MCID.  This works for my PDFs (see for example https://dhdaines.github.io/serafim/?idx=1057 and associated PDF behind the "PDF" button) but as noted above I need to re-test it with more interesting figures.

dhdaines:
> 2. Extracting a PDF's structure tree, which depends on `pypdfium2`, which introduces the overhead you mentioned, and appears to fail on Python 3.8 (something related to `_ctypes.PyCArrayType` but I haven't investigated closely).
> 
> Would it be _possible_ to separate this PR into those two distinct features? Is that something you'd be _open to_?

Absolutely, especially since part 2 should probably be rewritten to use `pdfminer.six`.

Actually part 1 ought to be a PR for `pdfminer.six` itself but it doesn't seem likely that it could be merged anytime soon.

dhdaines:
> 2. Extracting a PDF's structure tree, which depends on `pypdfium2`, which introduces the overhead you mentioned, and appears to fail on Python 3.8 (something related to `_ctypes.PyCArrayType` but I haven't investigated closely).

Ah, this can be fixed easily, it's just an issue with type-checking constructs that were introduced in Python 3.9 which slipped in accidentally.

pdfplumber/page.py
dhdaines: Yes, indeed, this was wrong.  I've fixed it to tag the lines/rects/curves with the MCID.

dhdaines:
Closing this PR and making two new ones!  (the MCID one is there already: #961 )

PR #946, opened 2023-07-24
Accept Iterable for geometry utils (fixes #945)

### LINKED ISSUE #945: `pdfplumber.utils` functions should take `Iterable` and not `List` arguments
Currently, `objects_to_bbox` and friends require a list as input, but there's no reason they couldn't work equally well with generator expressions, iterators, or anything else iterable.

Well, actually there is a reason, which is that they are written such that they iterate repeatedly over their input for each attribute (and thus can't reuse the same iterable), but this doesn't have to be the case, for instance, `merge_bboxes` could be written as:
```python
def merge_bboxes(bboxes: Iterable[T_bbox]) -> T_bbox:
    x0, top, x1, bottom = zip(*bboxes)
    return (min(x0), min(top), max(x1), max(bottom))
```
It's also a bit faster (note that there is no need to wrap the `map` in a `list`):
```
In [46]: %timeit pdfplumber.utils.merge_bboxes(list(map(bbox_getter, page.chars)))
464 µs ± 1.48 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
In [49]: %timeit merge_bboxes(map(bbox_getter, page.chars))
409 µs ± 1.18 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```
I can supply a PR :)

dhdaines:
This also makes bbox manipulation slightly faster as it no longer has to iterate 4 times over the same list.

Note: this removes the fragile-looking type manipulation in `intersects_bbox` as making it work with iterables is very ugly and I am not convinced of its usefulness (none of the other similar functions have it... I will let more experienced users/maintainers review and see if it is necessary!)

### DIFF
--- a/pdfplumber/utils/geometry.py
+++ b/pdfplumber/utils/geometry.py
@@ -1,52 +1,53 @@
 import itertools
 from operator import itemgetter
-from typing import Dict, List, Optional
+from typing import Dict, Iterable, Optional
 
 from .._typing import T_bbox, T_num, T_obj, T_obj_list
 from .clustering import cluster_objects
-from .generic import to_list
 
 
-def objects_to_rect(objects: T_obj_list) -> Dict[str, T_num]:
-    return {
-        "x0": min(map(itemgetter("x0"), objects)),
-        "x1": max(map(itemgetter("x1"), objects)),
-        "top": min(map(itemgetter("top"), objects)),
-        "bottom": max(map(itemgetter("bottom"), objects)),
-    }
+def objects_to_rect(objects: Iterable[T_obj]) -> Dict[str, T_num]:
+    """
+    Given an iterable of objects, return the smallest rectangle (i.e. a
+    dict with "x0", "top", "x1", and "bottom" keys) that contains them
+    all.
+    """
+    return bbox_to_rect(objects_to_bbox(objects))
 
 
-def objects_to_bbox(objects: T_obj_list) -> T_bbox:
-    return (
-        min(map(itemgetter("x0"), objects)),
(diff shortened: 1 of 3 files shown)


jsvine:
Thanks @dhdaines! This broadly sounds like a good change. Looks like the linter is angry, though. Could you run this locally, and then adjust until `make format lint tests` fully passes?:

```
pip install -r requirements-dev.txt
make format lint tests
```


> Note: this removes the fragile-looking type manipulation in intersects_bbox as making it work with iterables is very ugly and I am not convinced of its usefulness (none of the other similar functions have it... I will let more experienced users/maintainers review and see if it is necessary!)

Ah, thanks for flagging! I think this might be a relic of some old code, when I was trying to let all the utils accept pandas DataFrames in addition to lists of dicts.

dhdaines:
Hmm, I could have sworn that `make lint` and `make tests` passed, it could be that I have the wrong versions of `isort` and `black`...

There's a problem, at least for me, which is that `mypy` does not like the recent version of `numpy` that gets installed under Python 3.10:
```
venv/bin/python -m mypy --strict --implicit-reexport pdfplumber
venv/lib/python3.10/site-packages/numpy/__init__.pyi:638: error: Positional-only parameters are only supported in Python 3.8 and greater
```
This turns out to actually be a fixed bug in `mypy` (https://github.com/python/mypy/issues/13627), so I've updated `requirements-dev.txt` to have the version of `mypy` with the fix.

jsvine (approved):
Thanks again for this @dhdaines. And thanks for adding those docstrings. Really appreciate the thoroughness.

PR #961, opened 2023-08-09
Support for marked content section IDs

dhdaines:
As requested, this is the MCID part of #937 split out.  Structure tree support (using `pdfminer.six`) will be a separate PR.


dhdaines:
Note!  This page only extracts *marked-content identifiers* for sequences of objects.  There ~are a few other kinds~ is one kind of marked content that exist in PDF which it doesn't handle:

- ~marked-content sequences with tags (and no identifiers) - This PR will be fixed ASAP to support this~ DONE!
- marked-content *points* - These are marked points (with a tag and possible attributes) in the content stream which don't correspond to any given object.  It isn't clear how this could be supported in `pdfplumber`.

jsvine:
Many thanks for this, @dhdaines! It's a clever solution, and adds what seems like will be a powerful feature for people working with PDFs that have marked content. 

For now, I'm going to mark `mcid` and `tag` in the README as experimental attributes, but will remove that note if/when the `pdfminer.six` internals that make this possible remain stable.

dhdaines:
> Many thanks for this, @dhdaines! It's a clever solution, and adds what seems like will be a powerful feature for people working with PDFs that have marked content.
> 
> For now, I'm going to mark `mcid` and `tag` in the README as experimental attributes, but will remove that note if/when the `pdfminer.six` internals that make this possible remain stable.

Thank you! I will submit another PR soon to add the tag attributes, as these are useful for identifying headers and footers.

PR #965, opened 2023-08-10
fix issue 964

### LINKED ISSUE #964: extracted word is broken

```
page.extract_text()
```

## PDF file
[example.pdf](https://github.com/jsvine/pdfplumber/files/12310356/example.pdf)

## Expected behavior

extracted line:

`VLHDU8SHRR Homeowner Discount .....`

## Actual behavior

`VLHDU8SHRR H o m e o w ner Discount  .....`

## Screenshots

## Environment

- pdfplumber version: [0.10.2]
- Python version: [e.g., 3.8.1]
- OS: [e.g., Mac]

## Additional context

using param `use_text_flow=True` will avoid this bug, but this param will cause other extract format bugs like:

expected: `foo:  bar`
actual: `bar foo:`

jnhyperion:
I found that this issue is caused by some blank chars is overlapped with the following non blank chars.
The simple solution is to remove these overlapped blank chars.

fix: https://github.com/jsvine/pdfplumber/issues/964


jsvine:
Thanks for this proposal, @jnhyperion. I think this particular change isn't quite right for the library, as it's quite specific to a particular (and relatively uncommon) edge case. I find that changes like those might fix the handling of *some* PDFs, but risk causing problems for others, as there's such a wide variety of PDFs. But perhaps we can think of a more general feature that would still help for your use case, such as a simple `.extract_text(ignore_whitespace=True)` parameter or `Page.remove_whitespace(..., only_overlapping=True)` method (in a similar spirit to `Page.dedupe_chars(...)`).

jnhyperion:
Make sense, feel free to close this PR.

jnhyperion:
you're right, I added a new method `Page.remove_whitespace`.

PR #967, opened 2023-08-10
Add --structure-text flag to CLI (like `pdfinfo -struct-text` but better)

dhdaines:
This depends on my two other PRs (#961 and #963) so don't merge it before them!

Produces helpful JSON output, much like `pdfinfo  -struct-text`, but better, since it's actual JSON and not arbitrary text (or arbitrary XML, but we could do that too if you think it's useful)


dhdaines:
I will rebase this (and force-push) to make it cleaner to merge and fix the conflicts.

dhdaines:
Actually, no, I will just incorporate it into #963 after rebasing that one onto the current develop branch!

dhdaines:
Merged into #963

PR #1041, opened 2023-11-04
Add `x_tolerance_ratio` param to `extract_text` and similar functions (now properly linted!)

### LINKED ISSUE #987: For text extraction, add fractional versions of `x/y_tolerance` arguments
Currently, `x_tolerance` and `y_tolerance` are treated as numeric constants. But, as @Sarke points on in https://github.com/jsvine/pdfplumber/issues/606#issuecomment-1703456276, it could be useful to provide a "fractional" version of these arguments:

> [...] ideally we would be able to set the tolerance as a fraction of the font-size, since both the words spacing and the line spacing usually change proportionally with the change in font-size.

Implementing this correctly might be tricky, as `x/y_tolerance` are passed across a few methods, but it should be doable. Some other things to sort out:

- Let's say the new x-related parameter is `x_tolerance_fraction`. Should the value be a number (indicating desired fractional threshold) or a boolean (indicating that `x_tolerance` should be *interpreted* as a fraction)?
- What should the precise calculation be? Something like (in pseudocode) `x_tolerance = current_character["size"] * x_tolerance_fraction`?

Any other questions or complications I may be overlooking?

afriedman412:
Fix https://github.com/jsvine/pdfplumber/issues/987 (partially)

Passing x_tolerance_ratio to extract_text() and any other function that relies on WordExtractor will use the ratio * text size to determine where words begin and end. Overrides the x_tolerance param.

There is room to build out y_tolerance_ratio too, if need be in the future!


jsvine:
Thanks, and now merging!

PR #1094, opened 2024-02-14
Handle missing ParentTree

dhdaines:
It's not supposed to happen according to section 14.7.4.4 of the PDF 1.7 standard:

> The parent tree is a number tree (see 7.9.7, “Number Trees”), accessed from the ParentTree entry in a
document’s structure tree root (Table 322). The tree shall contain an entry for each object that is a content item
of at least one structure element and for each content stream containing at least one marked-content sequence
that is a content item. The key for each entry shall be an integer given as the value of the StructParent or
StructParents entry in the object (see Table 326)

But of course it does, for example in https://assets-global.website-files.com/5e691d0b7de02f1fd6919876/65b7f08c45d632b34f027ec9_Mines%20Against%20Humanity%20EN.pdf

Handle this gracefully with the "unparsed pages" mechanism.

### DIFF
--- a/pdfplumber/structure.py
+++ b/pdfplumber/structure.py
@@ -86,24 +86,30 @@ class PDFStructTree:
         # If we have a specific page then we will work backwards from
         # its ParentTree - this is because structure elements could
         # span multiple pages, and the "Pg" attribute is *optional*,
-        # so this is the approved way to get a page's structure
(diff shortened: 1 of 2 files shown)


jsvine:
Thanks! Merging into `develop`.

PR #1095, opened 2024-02-14
Some utility methods for logical structure

dhdaines:
It's useful to be able to search in the structure tree - this has to be done from the `PDFStructTree` object itself since we return a dictionary from `structure_tree` in keeping with the general way of `pdfplumber`.

Also to get a BBox from an element for  visual debugging - note the FIXME, if you play games with cropped pages, this will fail, but in general that's unlikely, you would have to do something like:

```python
pdf = pdfplumber.open(pdffile)
page = pdf.pages[0].crop(some_bbox)
stree = PDFStructTree(pdf, page)  # NO! Don't do this! Why would you do this?
```
and then try to get the BBox of an element where it is explicitly specified in the attributes of that element (usually this is only the case for `Figure` and `Table`).

Is there a general method to properly transform PDF BBoxes into `pdfplumber` ones for a page?

### DIFF
--- a/docs/structure.md
+++ b/docs/structure.md
@@ -59,3 +59,18 @@ In this case, because marked content IDs are specific to a given page,
 each element will also have a `page_number` attribute, which is the
 number of the page containing (partially or completely) this element,
 indexed from 1 (for consistency with `pdfplumber.Page`).
+
+You can also access the underlying `PDFStructTree` object for more
+flexibility, including visual debugging.  For instance to plot the
+bounding boxes of the contents of all of the `TD` elements on the
+first page of a document:
+
+    page = pdf.pages[0]
+    stree = PDFStructTree(pdf, page)
+    img = page.to_image()
+    img.draw_rects(stree.element_bbox(td) for td in table.find_all("TD"))
+
+The `find_all` method works rather like the same method in
+[BeautifulSoup](https://beautiful-soup-4.readthedocs.io/en/latest/#searching-the-tree) -
+it takes an element name, a regular expression, or a matching
+function.
--- a/pdfplumber/structure.py
+++ b/pdfplumber/structure.py
@@ -1,7 +1,19 @@
 import logging
+import re
 from collections import deque
 from dataclasses import asdict, dataclass, field
-from typing import TYPE_CHECKING, Any, Dict, Iterator, List, Optional, Tuple
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Dict,
+    Iterable,
+    Iterator,
+    List,
+    Optional,
+    Pattern,
+    Tuple,
+)
 
 from pdfminer.data_structures import NumberTree
 from pdfminer.pdfpage import PDFPage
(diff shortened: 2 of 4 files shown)


docs/structure.md
```diff
@@ -59,3 +59,18 @@ In this case, because marked content IDs are specific to a given page,
...
+
+You can also access the underlying `PDFStructTree` object for more
+flexibility, including visual debugging.  For instance to plot the
+bounding boxes of the contents of all of the `TD` elements on the
+first page of a document:
+
+    page = pdf.pages[0]
+    stree = PDFStructTree(pdf, page)
+    img = page.to_image()
+    img.draw_rects(stree.element_bbox(td) for td in table.find_all("TD"))
```
jsvine, line 71: Neat!

jsvine:
I haven't played around with this yet, but it seems like a reasonable idea and doesn't interfere with core `pdfplumber` functionality, so I'm inclined to merge. Is it ready to merge?

> Is there a general method to properly transform PDF BBoxes into `pdfplumber` ones for a page?

If I'm understanding correctly, this this question pertains to flipping the vertical coordinates, so that `(x0, y0, x1, y1)` (i.e., bbox with origin at the bottom-left) becomes `(x0, top, x1, bottom)` (origin at top-left). Is that right? If so: We just calculate the `top` and `bottom` attributes once, on parsing:

https://github.com/jsvine/pdfplumber/blob/1ad3905612d6c9ac9b285332850da23a0a96f0ba/pdfplumber/page.py#L399-L402

... and then use `(x0, top, x1, bottom)` as the standard bbox throughout.

dhdaines:
> I haven't played around with this yet, but it seems like a reasonable idea and doesn't interfere with core `pdfplumber` functionality, so I'm inclined to merge. Is it ready to merge?

I think maybe I'll add a companion / convenience method `find` to get just the first instance of an element, and at least minimally handle the issue below (it's a bit more complicated, basically I can use the example code above as a unit test).

> > Is there a general method to properly transform PDF BBoxes into `pdfplumber` ones for a page?
> 
> If I'm understanding correctly, this this question pertains to flipping the vertical coordinates, so that `(x0, y0, x1, y1)` (i.e., bbox with origin at the bottom-left) becomes `(x0, top, x1, bottom)` (origin at top-left). Is that right? If so: We just calculate the `top` and `bottom` attributes once, on parsing:

The issue is a bit more complicated because when you crop a page, all of the object coordinates go through the `crop_fn` which adjusts them.  So far, so good, but structure tree elements can have a `BBox` attribute specified on them which is not attached to any particular object.  In `element_bbox` we will prefer this if it exists, but then we will also have to transform it, not only flipping the vertical coordinates, but also applying `crop_fn`.

I can probably hack something up so this minimally works but we might want to refactor it at some point.

dhdaines:
Should be ready to merge now!

I didn't realize that cropping a page doesn't actually translate the coordinates of the objects, it just clips them to the new bounding box - nonetheless, this didn't work right for structure elements with BBox attributes, and now it does.

docs/structure.md
dhdaines: now you can also do `table = stree.find("Table")` to get a single element.

jsvine:
Thanks, now merged! (And correct re. the non-translation of coordinates.)

Issue #1114, opened 2024-03-14
Custom deduppe_chars char properties


felix-hh:
Following up on  #71  - I have had a problem with duplicated characters that have the same text but different properties (i.e. fonts). Unfortunately I can't share the file as it is private. I am requesting optionally providing custom properties to the deduplication function (please let me know if this is otherwise available!)

Here's a sketch of the proposed code changes:

https://github.com/jsvine/pdfplumber/blob/147f2c4c07dc1191fc1d05bb589b4f6af3aaf74a/pdfplumber/utils/text.py#L789

```python
def dedupe_chars(chars: T_obj_list, tolerance: T_num = 1, char_properties: Optional[List[str]] = None) -> T_obj_list:
    """
    Removes duplicate chars — those sharing the same text, fontname, size,
    and positioning (within `tolerance`) as other characters in the set.
    """
    # key = itemgetter("fontname", "size", "upright", "text")
    char_properties = char_properties if char_properties is not None else ["fontname", "size", "upright", "text"]
    key = itemgetter(*char_properties)

   <... more code>
```

The interfaces exposing this should also be updated. 

The end result looks like 
```python
print(section.dedupe_chars(tolerance=0.1, char_properties=['text']).extract_text())
```

jsvine:
Thanks for the suggestion, @felix-hh. Are you able to share a version of the PDF redacted with https://github.com/JoshData/pdf-redactor? Or another PDF that demonstrates the same issue?

felix-hh:
Hi @jsvine I made a good-faith attempt at redacting the pdf with the tool but the footer text is not redacted and can still be extracted. This is a problem because the footer identifies the data source which is proprietary. I also do not know how to reproduce the issue with my own pdf.

felix-hh:
Let me know if there is some other way I can help. I am happy to provide a pull request for the change verifying that it works on my end. 

Here's some screenshots if it helps:
**Redacted PDF screenshot:**
![image](https://github.com/jsvine/pdfplumber/assets/50616313/39977831-5b47-4ce5-9504-94e77757d10b)


**What the output of extract_text looks like:**
![image](https://github.com/jsvine/pdfplumber/assets/50616313/e0038895-3816-4c76-94b0-f303b80aee0f)

jsvine:
[...]

PR #1161, opened 2024-06-27
add ignore_char_properties arg in dedupe_chars

QuentinAndre11:
See [#1114](https://github.com/jsvine/pdfplumber/issues/1114)

More precise parameters could be used for char matching in dedupe_chars (a list of list of similar fontnames, a tolerance for the size...) but for my use case directly ignoring these keys was sufficient and it was easier to test.

PS: It is my first contribution so I did not know if I had to add a new version in the Changelog or append it to the last one.


jsvine:
Hi @QuentinAndre11, and thanks for your interest in `pdfplumber`! I'm curious: What advantage do you see to this approach vs. having the list of properties be directly configurable? E.g., instead of this in the current `pdfplumber` code:

``` python
def dedupe_chars(chars: T_obj_list, tolerance: T_num = 1) -> T_obj_list:
    """
    Removes duplicate chars — those sharing the same text, fontname, size,
    and positioning (within `tolerance`) as other characters in the set.
    """
    key = itemgetter("fontname", "size", "upright", "text")
```

... doing this:

```python
def dedupe_chars(
  chars: T_obj_list, 
  tolerance: T_num = 1, 
  properties: Tuple[str, ...] = tuple("fontname", "size", "upright", "text")
) -> T_obj_list:
    """
    Removes duplicate chars — those sharing the same text, fontname, size,
    and positioning (within `tolerance`) as other characters in the set.
    """
    key = itemgetter(*properties)
```

QuentinAndre11:
Hi ! I thought text and upright properties aren't properties that we would want to change at all. The only case I can think of for the text would be visually close letters like I and l but why on earth would we want to deduplicate that ? Same goes for upright-proof letters like o but it also does not seem a relevant feature.
So because we have mandatory properties, it made more sense for me to "tolerate" some props than to directly give the keys. Does it make sense ?

Ps: the tests are failing because of pandas, I guess I can't do anything about it?

QuentinAndre11:
If needed, I can actually replace the kwargs from dedupe_chars(tolerate) to dedupe_chars(tolerate_x: T_num = 1, tolerate_y: T_num = 1, ignore_font: bool = False, tolerate_size: T_num = 0), it's not that much additional work @jsvine 
Edit: we need to keep tolerate for backward compatibility, so tolerate: Union[T_num, Tuple[T_num, T_num]] = 1

jsvine:
Thanks, @QuentinAndre11. This is a good point:

> I thought text and upright properties aren't properties that we would want to change at all.

Given that there's no real reason to change them, what about this updated definition?:

```python
def dedupe_chars(
  chars: T_obj_list, 
  tolerance: T_num = 1, 
  extra_attrs: Optional[Tuple[str, ...]] = tuple("fontname", "size")
) -> T_obj_list:
    """
    Removes duplicate chars — those sharing the same text, fontname, size,
    and positioning (within `tolerance`) as other characters in the set.
    """
    key = itemgetter(*("upright", "text"), *(extra_attrs or tuple()))
```

Then, all you'd have to do for your particular use-case is call `.dedupe_chars(extra_attrs=None)`. At the same time, it would enable folks to be *even more strict*, if they preferred, by passing a more extensive list of attributes.

Another advantage is that the parameter matches the name (and idea) of the `.extract_words(extra_attrs=...)`.

QuentinAndre11:
@jsvine Works for me (and my use case), so I implemented it your way !

jsvine:
Thanks! Pushed some documentation tweaks and a fix for the failing tests (we just needed to upgrade `pandas`).

jsvine:
Looks like there's still an issue with Python 3.8 (because the latest `pandas` isn't available on it), but that version is [reaching its end of life](https://endoflife.date/python), so we'll remove 3.8 from the test suite after merging. Thanks again for the contribution!

PR #1169, opened 2024-07-10
fix: dereference structure element attributes

dhdaines:
It may or may not be allowed in the PDF specification, but attributes of structure objects are sometimes references instead of simple strings/symbols.  In that case structure tree extraction will crash in an unfortunate way.  So this makes sure to dereference them if necessary!


jsvine:
Roger that, and thanks for describing the reasoning 👍

Issue #1159, opened 2024-06-23
Error While extracting Non-English tables (Arabic) reversed Text


iiAmeer:
I have a problem, when i use extract_tables function, i get a reversed arabic text. Like the actual name is joe, but i get eoj. (it's arabic but i wrote english so u can understand), there is no need for screenshots its clear.

## Environment

- pdfplumber version: latest 
- Python version: 3.11
- OS: Windows (later linux)

jsvine:
Thanks for noting this. Could you share the PDF? That will make it easier to diagnose the issue and suggest solutions.

iiAmeer:
> Thanks for noting this. Could you share the PDF? That will make it easier to diagnose the issue and suggest solutions.

![087_ثانوية الإمام علي بن ابي طالب الأهلية للبنات (1).pdf](https://github.com/user-attachments/files/15975408/087_.1.pdf)

This is the file i used, if you want my code & run preview wait me until tomorrow . 

![IMG_20240626_092555.jpg](https://github.com/jsvine/pdfplumber/assets/68138224/a51a51eb-f0b2-4b62-86f7-aafebffd7a1e)

jsvine:
Thanks for providing the PDF. Ignoring the numbers for now, and instead focusing on the text ... does the output of this look more like what you want?:

```python
import pandas as pd
table = page.extract_table(dict(text_char_dir_render="rtl"))
pd.DataFrame(table, columns=None).fillna("")
```
<img width="854" alt="Screenshot 2024-07-01 at 11 09 20 PM" src="https://github.com/jsvine/pdfplumber/assets/534702/000b3004-9d3e-45a8-93b2-7e9532fdac06">

iiAmeer:
> Thanks for providing the PDF. Ignoring the numbers for now, and instead focusing on the text ... does the output of this look more like what you want?:
> 
> ```python
> import pandas as pd
> table = page.extract_table(dict(text_char_dir_render="rtl"))
> pd.DataFrame(table, columns=None).fillna("")
> ```
> 
[...]

PR #1187, opened 2024-08-15
Fixing RTL little Bug 

iiAmeer:
i've Fixed issue #1159 that opend by me (:


jsvine:
Hi @iiAmeer, and thank you for your continued interest in this topic. Although the proposed change might fix the output for the particular PDF(s) you're working with, I think the RTL issue is somewhat more complex, based on other example PDFs I've examined.

PR #1195, opened 2024-08-30
Feature: Warn on unicode decoding errors in PDF annotations

stolarczyk:
In certain scenarios, annotations may include invalid or extraneous data that can obstruct the annotation processing workflow. To mitigate this, the `warn_unicode_error` parameter in the `PDF` initializer and the `.open()` method provides a configurable option to bypass these errors and generate warnings instead, ensuring smoother handling of such anomalies. 

Example warning, if activated:

```
UserWarning: Could not decode contents for annotation. Annotation contents will be missing.
```

### DIFF
--- a/pdfplumber/page.py
+++ b/pdfplumber/page.py
@@ -312,7 +312,10 @@ class Page(Container):
                         except UnicodeDecodeError:
                             if self.pdf.raise_unicode_errors:
                                 raise
-                            warn(f"Could not decode {k} for annotation. {k} will be missing.")
+                            warn(
+                                f"Could not decode {k} for annotation."
+                                " {k} will be missing."
+                            )
 
             parsed = {
                 "page_number": self.page_number,
--- a/pdfplumber/page.py
+++ b/pdfplumber/page.py
@@ -313,8 +313,8 @@ class Page(Container):
                             if self.pdf.raise_unicode_errors:
                                 raise
                             warn(
-                                f"Could not decode {k} for annotation."
-                                " {k} will be missing."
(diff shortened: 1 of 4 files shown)


jsvine:
Hi @stolarczyk, and thank you for this suggestion. It makes sense to provide such warnings, although I'd lean toward a more generalizable approach rather than specifying parameters for each type of warning. To that end, I'm more inclined to use Python's built-in [warning filtering](https://docs.python.org/3/library/warnings.html). I'm open to other opinions, though. What do you think?

stolarczyk:
thanks for looking into this, @jsvine!

I’m not entirely clear on how your idea regarding built-in warning filtering would address the issue I'm focusing on. The proposed change turns an _exception_ (`UnicodeDecodeError`) into a warning, which prevents the PDF processing from crashing entirely. So the warning is the result, not the issue at hand.

jsvine:
My apologies for the misunderstanding! I think the name of the proposed parameter threw me off, but I also should have looked more closely. I think I understand it now. This proposal makes sense, though what about tweaking the name?: `raise_unicode_errors=True/False`?

stolarczyk:
Thanks for the suggestion. Just renamed it.

jsvine:
Thanks, @stolarczyk — I've pushed a small tweak, above, so that the linter is happy. But looks like we're missing a bit of test coverage:

stolarczyk:
@jsvine, the test has been added.

jsvine:
Thanks! Merged into `develop`

PR #1235, opened 2024-12-10
Add a --format text option

brandonrobertz:
This PR adds a `--format text` option to the CLI interface.

I use this regularly because pdfplumber has among the best layout preserving methods for PDFs, especially machine generated ones. Exposing the page output via CLI lets me use pdfplumber as a general purpose PDF-to-text tool.

```
Usage:
pdfplumber --format text file.pdf > file.txt
```

If there's some other way I should perform this (other than `print(extract_text(...))` LMK and I'll make changes.

### DIFF
--- a/pdfplumber/cli.py
+++ b/pdfplumber/cli.py
@@ -37,7 +37,7 @@ def parse_args(args_raw: List[str]) -> argparse.Namespace:
         action="store_true",
     )
 
-    parser.add_argument("--format", choices=["csv", "json"], default="csv")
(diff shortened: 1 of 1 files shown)


jsvine:
Thanks! And glad to hear `pdfplumber` has been useful. I think this makes a lot of sense as a feature. Future iterations could allow passing relevant config params, such as `--x-density`, `--y-density`, `--x-tolerance`, `--y-tolerance`, et cetera. But even this config-less version seems worth including.

Issue #1089, opened 2024-02-05
`Page.to_image()` leaks file descriptors


dhdaines:
## Describe the bug

When extracting images from a PDF file with a large number of pages, we eventually run out of file descriptors on Mac.  This probably leaks file descriptors everywhere but Linux has less strict ulimits.  It may also be related to #1072 as Windows has silly restrictions on opening the same file multiple times.

This is because `pypdfium2` is holding onto a file descriptor somewhere.  The exception thrown by the code below:

```
Traceback (most recent call last):
  File "/Users/dhd/work/vdsa/alexi/download/pdfbug.py", line 5, in <module>
    img = page.to_image()
          ^^^^^^^^^^^^^^^
  File "/Users/dhd/work/vdsa/pdfplumber/pdfplumber/page.py", line 585, in to_image
    return PageImage(
           ^^^^^^^^^^
  File "/Users/dhd/work/vdsa/pdfplumber/pdfplumber/display.py", line 88, in __init__
    self.original = get_page_image(
                    ^^^^^^^^^^^^^^^
  File "/Users/dhd/work/vdsa/pdfplumber/pdfplumber/display.py", line 56, in get_page_image
    pdfium_page = pypdfium2.PdfDocument(
                  ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dhd/work/vdsa/venv/lib/python3.12/site-packages/pypdfium2/_helpers/document.py", line 62, in __init__
    input = input.expanduser().resolve()
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/pathlib.py", line 1240, in resolve
    s = self._flavour.realpath(self, strict=strict)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen posixpath>", line 436, in realpath
  File "<frozen posixpath>", line 423, in abspath
OSError: [Errno 24] Too many open files
```

## Code to reproduce the problem

```python
import pdfplumber
with pdfplumber.open("20231213-Codification-administrative-Rgl-1314-2021-Z.pdf") as pdf:
    for page in pdf.pages:
        img = page.to_image()
```

## PDF file

https://ville.sainte-adele.qc.ca/upload/documents/20231213-Codification-administrative-Rgl-1314-2021-Z.pdf (485 pages, enough to hit the ulimit on a Mac!)

[...]

PR #1270, opened 2025-02-02
test_issue_1089: update wording regarding pypdfium2

mara004:
pypdfium2 doesn't actually "leak" file descriptors.
It's just that if the caller doesn't close the PdfDocument explicitly, we rely on garbage collection / finalization to do that, which may happen too late. Update the wording accordingly.

See also https://github.com/jsvine/pdfplumber/issues/1089#issuecomment-2007961627 for background.

### DIFF
--- a/tests/test_issues.py
+++ b/tests/test_issues.py
@@ -285,10 +285,11 @@ class Test(unittest.TestCase):
 
     def test_issue_1089(self):
         """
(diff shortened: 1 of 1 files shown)


jsvine:
Thanks for the clarification! Merged into `develop`.

PR #1285, opened 2025-03-26
Fix pdfminer.six dependencies

pprados:
The latest version of `pdfminer.six` had issues with certain PDF files.  

The pdfplumber code was forced to declare a dependency on an older version: `pdfminer.six==20231228`.  
This caused problems when trying to use multiple different parsers, with different versions, in the same project.  

I [fixed the bug](https://github.com/pdfminer/pdfminer.six/pull/1081) in `pdfminer.six`. The [latest released version](https://pypi.org/project/pdfminer.six/) includes this modification, so it is no longer necessary to depend on an older version.  

This fix updates the dependency on `pdfminer.six` to remove this limitation.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
-pdfminer.six>=20240706
+pdfminer.six>=20250324
 Pillow>=9.1
 pypdfium2>=4.18.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
-pdfminer.six==20231228
+pdfminer.six>=20240706
 Pillow>=9.1
 pypdfium2>=4.18.0


walterheck:
@pietermarsman it would be really great to have this merged and released so the dependencies get unstuck

jsvine:
Just pushed `pdfplumber==0.11.6`, which updates the pinned `pdfminer.six` version. This dependency, however, will generally remain pinned, since `pdfplumber` depends on parts of `pdfminer.six` that are not guaranteed (nor necessarily expected) to be stable over time.

jsvine:
p.s., thank you for the `pdfminer.six` fix you mention above, @pprados!

PR #1295, opened 2025-05-02
Minor cleanup around pypdfium2 integration

mara004:
Avoid accessing a public member through a private path.
Also, remove a misleading comment (the `autoclose` option is about byte stream input, i.e. unrelated).

This is basically a follow-up on #1270.

### DIFF
--- a/pdfplumber/display.py
+++ b/pdfplumber/display.py
@@ -55,7 +55,7 @@ def get_page_image(
 
     try:
         pdfium_doc = pypdfium2.PdfDocument(src, password=password)
-    except pypdfium2._helpers.misc.PdfiumError as e:
+    except pypdfium2.PdfiumError as e:
         raise MalformedPDFException(e)
 
     pdfium_page = pdfium_doc.get_page(page_ix)
@@ -69,8 +69,6 @@ def get_page_image(
         # Non-modifiable arguments
         prefer_bgrx=True,
     ).to_pil()
-    # In theory `autoclose` when creating it should make it close...
-    # automatically.  In practice this does not seem to be the case.
     pdfium_doc.close()
 
     return img.convert("RGB")


pdfplumber/display.py
```diff
@@ -69,8 +69,6 @@ def get_page_image(
         # Non-modifiable arguments
         prefer_bgrx=True,
     ).to_pil()
-    # In theory `autoclose` when creating it should make it close...
-    # automatically.  In practice this does not seem to be the case.
     pdfium_doc.close()
 
     return img.convert("RGB")
```
mara004, line 74: Oh, by the way, why are you passing `prefer_bgrx=True` but then convert to `RGB` using PIL ?
By omitting this option, you'd avoid the need to convert. `RGB` should be the default mode you'll get with `page.render(...).to_pil()`.

jsvine: If I recall correctly, some users were seeing strange conversion artifacts, and we landed on this solution. But it is entirely possible that I misunderstood / mis-solved.

jsvine:
Helpful on both counts, thank you!

pdfplumber/display.py
mara004: That sounds odd, do you have a pointer to these user reports?

BTW, you might also want to pass `rev_byteorder=True` to have (py)pdfium actually output RGB byte order. By default it outputs BGR but the PIL opener will remap to RGB.
That said, AFAIK, PIL's backend cannot actually work with RGB under the hood and would copy internally, so if you want a zero copy mutable view of the original buffer you would need to have (py)pdfium output RGBA or RGBx.
Maybe RGBA would be best, because pdfium is inefficient at rendering to {RGB,BGR}{x} in corner cases that involve transparency.

Sorry for the compliation.

jsvine: Thank you for this insight. I'm not terribly familiar with image conversion. Here are some of the relevant prior changes/discussions in case the context helps:

- https://github.com/jsvine/pdfplumber/commit/7e28931
- https://github.com/jsvine/pdfplumber/discussions/899#discussioncomment-6464765
- https://github.com/jsvine/pdfplumber/pull/340

mara004: I just skimmed the threads, but sounds mostly historical/unrelated to me, e.g. (3) was with `wand`, not pypdfium2.

I think it should be fine to drop `prefer_brgx=True` and the conversion.

PR #1300, opened 2025-05-19
patched to do --help if run wo args, instead of autistic screeching

PopeNobody:
It's so much more chill that way!


PopeNobody:
If you guys want it you got it.

jsvine:
That seems like a reasonable adjustment to the code. Merged. Thank you.

PR #1314, opened 2025-06-11
update PyMuPDF notes

mjhea0:


mjhea0:
Just wanted to add a note about the license for PyMuPDF. It's fairly restrictive:

> You cannot deploy our open-source in a network without disclosing your own application’s full source code under AGPL, regardless of whether the network is internal or external, with any users interacting with it.

Source: https://artifex.com/licensing

jsvine:
Thanks for flagging. Providing some indication of licensing constraints makes sense. In this case, though, I think I would prefer just to indicate the license for each of the mentioned tools (rather than only for PyMuPDF). If this is a change you're interested in making, great. If not, no worries; I can incorporate separately.

jsvine:
Thanks again, @mjhea0; licensing info added in [b67079f](https://github.com/jsvine/pdfplumber/commit/b67079fb70bf5eec0179efcf9f0e5d006b732749) above.

Issue #1325, opened 2025-07-20
Add logic to add outer borders to well-defined tables that lack them


jsvine:
See, e.g., [this table from #808](https://github.com/jsvine/pdfplumber/discussions/808#discussioncomment-13030277): 

<img width="746" height="512" alt="Image" src="https://github.com/user-attachments/assets/0d956dee-92e4-46f6-a7b2-a2d939664f3b" />

It seems possible that `TableFinder` could, after a first pass at detecting tables, "complete" the outer borders of tables like these. The first step might be looking for lines/rects that extend outside the detected cells.

u7923257-ANU:
Could this please be assigned to myself / [DoomGuy77](https://github.com/DoomGuy77)?

PR #1342, opened 2025-10-26
Feature: Detect cells without outer border when using the lines strategy for table detection

DoomGuy77:
This PR implements the feature discussed in #1325 by creating extra edges to close the borders of open tables. We found this easier to implement than @jsvine 's suggested approach.
We added an option in `TableOptions` for whether to use this to close tables

Note that this is still a work in progress


kalle07:
thx for working on it   ;)

PR #1353, opened 2026-01-22
Adjust text-strategy vertical edges to nearest word gap

coreyberla:
When using vertical_strategy="text", edges are placed at cluster boundaries which can fall inside wider words (e.g., a long header with short data below). This adjusts such edges to the midpoint of the nearest gap between words, preventing table columns from splitting text.

Example.  See the header "Expense" in column 5.

### Before
image

### After
image

### DIFF
--- a/pdfplumber/table.py
+++ b/pdfplumber/table.py
@@ -141,6 +141,57 @@ def words_to_edges_h(
     return edges
 
 
+def _find_nearest_gap_midpoint(edge_x: T_num, words: T_obj_list) -> T_num:
+    """
+    Find the midpoint of the nearest gap between words for an edge position.
+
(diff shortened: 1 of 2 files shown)


ancientwaiting:
Got your email, thank you!Will reply you asap

jsvine:
Thank you for interest in pdfplumber, and for suggesting improvements! I'm not entirely convinced that this adjustment leads to preferable results in most/all cases (as opposed just to more specific cases like those screenshotted above), but I'm open to being persuaded.

To that end: Could you include tests that operate on PDFs (see other tests in the `tests/` directory for examples), perhaps of a few different layouts?

And I'm curious: Was this code written in part or full by AI/LLM? If so, could you share more details about that approach?

PR #1354, opened 2026-01-25
Add extend_edges function to fix table extraction with one strat text and the other non-text

monchin:
Hi, I refered pdfplumber to write a library to extract pdf tables, and found a table-extracting bug when one strategy is "text" and the other is not, please see monchin/tablers#8 for more details.

I have fixed it in my library, and I found it also occurs in pdfplumber, so I'd like to fix it.


jsvine:
Thank you for suggesting this change, which sounds intriguing. I'll aim to take a closer look soon. And tablers looks like a neat library, thanks for sharing.

monchin:
> Thank you for suggesting this change, which sounds intriguing. I'll aim to take a closer look soon. And tablers looks like a neat library, thanks for sharing.

Thank you for your kind reply and admiration for `tablers`! 
I have formatted the code by `black==24.8.0`, and IMHO it would be great if  your `test.yml` workflow could be modified that could be triggered on PR on "develop" branch so that we can see the CI results without your approval.

PR #1364, opened 2026-03-18
Fix ResourceWarning leak for internally opened streams

### LINKED ISSUE #1336: Bug: Input file handle is never closed, causing a ResourceWarning in Python 3.11+
## Describe the bug

The input file handle doesn't appear to be closed properly in a variety of code paths, resulting in a ResourceWarning in modern Python versions. It would be great if this library automatically closed the handle properly.

Using context managers on file access is the best/proper way to solve this issue!

ReinerBRO:
## Summary
- register a finalizer to close internally-opened PDF streams if a PDF object is garbage collected without an explicit close
- keep externally provided streams untouched
- add a regression test covering the ResourceWarning leak and external-stream behavior

## Why
Issue #1336 reports that file handles opened by `pdfplumber.open(path)` can survive until garbage collection and emit a `ResourceWarning`. This change ensures internally owned streams are still closed if callers forget to use a context manager or call `close()`.

## Tests
- `. .venv/bin/activate && python -m pytest tests/test_resource_warning.py -q`

Closes #1336

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file. The format 
 ## Unreleased
 
 - Upgrade `pdfminer.six` from `20251230` to `20260107`. ([07a5ff6](https://github.com/jsvine/pdfplumber/commit/07a5ff6))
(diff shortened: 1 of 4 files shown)


ReinerBRO:
Closing this duplicate PR in favor of #1363, which carries the same commit and targets the intended branch for review.

PR #1371, opened 2026-05-13
Preserve explicit table lines outside the natural edge span (fixes #1335)

### LINKED ISSUE #1335: Table extraction fails when last column is misaligned or multi-line using vertical_strategy="explicit"
## Describe the bug

When using `extract_table()` with `vertical_strategy="explicit"` and `horizontal_strategy="text"`, the last column of a table is often missed or misinterpreted—especially when the column is narrow or contains multi-line text. This leads to incomplete or split rows in the output.

Yes, tested with `repair=True`. The issue persists.

```python
import pdfplumber

with pdfplumber.open("sample.pdf", repair=True) as pdf:
    page = pdf.pages[0]
    table = page.extract_table({
        "vertical_strategy": "explicit",
        "horizontal_strategy": "text",
        "intersection_tolerance": 5,
        "snap_tolerance": 3,
        "join_tolerance": 5
    })
    for row in table:
        print(row)

### PR DESCRIPTION
**Pull request by Cyberfilo**
## What

When `page.extract_table(...)` is called with `vertical_strategy="explicit"` (or `horizontal_strategy="explicit"`), the user-supplied lines define the column / row boundaries.  Edges of the opposite orientation are still derived from text clustering or detected page lines, and those are only as wide as the underlying objects.

`words_to_edges_h` (and the `lines` variants for horizontal edges) span just `min_x0..max_x1` of the rows they detect.  `edges_to_intersections` then keeps only the vertical edges whose `x0` falls inside `[h.x0 - x_tol, h.x1 + x_tol]`.  When an explicit vertical line sits outside that span — common when the leftmost or rightmost column is narrow, right-aligned, or contains only short / multi-line text — the intersection is rejected and the entire column collapses into its neighbour (or disappears entirely if it is the only one beyond the span).

## Fix

In `TableFinder.get_edges()`, after the base + explicit edges are concatenated, extend each horizontal edge's `x0` / `x1` to cover the x-range of any explicit vertical lines (and symmetrically extend vertical edges' `top` / `bottom` to cover any explicit horizontals).

Explicit edges already span the full page when supplied as floats, and span their declared bbox when supplied as dicts — so this only ever **widens** edges derived from words or page lines, never narrows or moves anything.  Tables that do not use explicit lines are untouched.

## Repro / test

Added `test_issue_1335_explicit_outside_text_span` in `tests/test_table.py`.  It constructs a three-row, three-column layout with a narrow last column (text ends ~130 px before the rightmost explicit vertical) and asserts that:

- all four explicit vertical positions produce intersections;
- the resulting table contains rows with all three column-cells present.

Without the patch, only the inner two explicit verticals produce intersections and the table collapses to a single column.  Running the new test against `develop`:

```
$ pytest tests/test_table.py::Test::test_issue_1335_explicit_outside_text_span
... FAILED  (assert [200.0, 350.0] == [50, 200, 350, 500])
```

With the patch:

```
$ pytest tests/test_table.py::Test::test_issue_1335_explicit_outside_text_span
... PASSED
```

## Verification

- `make tests` — 165 passed (the pre-existing `test_repair.py::test_from_issue_932` failure on systems without Ghostscript is unchanged).
- `make lint` — clean (flake8, mypy --strict, black --check, isort --check).
- Added entry under `## Unreleased` in `CHANGELOG.md` and added myself to the contributors list per `CONTRIBUTING.md`.

Fixes #1335.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file. The format 
 
 ## Unreleased
 
+### Fixed
+- Fix `extract_table` dropping columns when `vertical_strategy="explicit"` is paired with text- or lines-derived horizontal edges that don't span the full range of the explicit vertical lines (and symmetrically for explicit horizontal lines). ([#1335](https://github.com/jsvine/pdfplumber/issues/1335))
+
+### Changed
 - Upgrade `pdfminer.six` from `20251230` to `20260107`. ([07a5ff6](https://github.com/jsvine/pdfplumber/commit/07a5ff6))
 
 ## 0.11.9 — 2026-01-05
--- a/README.md
+++ b/README.md
@@ -574,6 +574,7 @@ Many thanks to the following users who've contributed ideas, features, and fixes
 - [Brandon Roberts](https://github.com/brandonrobertz)
 - [@ennamarie19](https://github.com/ennamarie19)
 - [Anton Ilin](https://github.com/bronislav)
+- [Filippo Mattia Menghi](https://github.com/Cyberfilo)
 
 ## Contributing
 
--- a/pdfplumber/table.py
+++ b/pdfplumber/table.py
@@ -694,6 +694,31 @@ class TableFinder(object):
 
         h = h_base + h_explicit
 
+        # Make sure horizontal edges span the x-range of any explicit vertical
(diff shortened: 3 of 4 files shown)


Cyberfilo:
@jsvine @samkit-jain please review

PR #1372, opened 2026-05-22
Fix `to_image()` omitting filled AcroForm widget values (closes #1367)

### LINKED ISSUE #1367: `to_image()` / `get_page_image` omits filled AcroForm widgets because `PdfDocument.init_forms()` is never called
## Describe the bug

Rasterizing with `page.to_image()` goes through `pdfplumber.display.get_page_image`, which opens a `pypdfium2.PdfDocument` and loads the page without calling **`PdfDocument.init_forms()`**. Filled **AcroForm** field content is drawn via PDFium’s form layer (`FPDF_FFLDraw`); in **pypdfium2** that only runs when a form environment exists, which **`init_forms()`** creates (and must be called after open, **before** loading pages).

So **filled form field text can be missing from the PIL bitmap** while PDF viewers still show it. This is **PDFium rendering**, not pdfminer **text extraction** (e.g. `.chars`).

Yes. `pdfplumber.open(..., repair=True)` does not fix this: the PDF is not treated as malformed for this path. The gap is that **`get_page_image`** never initializes the PDFium form environment before `render()`.

```python
import pdfplumber

with pdfplumber.open("filled_form.pdf") as pdf:
    im = pdf.pages[0].to_image(resolution=150).original
    im.save("out.png")
```

(Also reproducible with `repair=True` on the same file.)

## PDF file

[filled_form.pdf](https://github.com/user-attachments/files/26534899/filled_form.pdf)

## Expected behavior

`out.png` should include the **visible filled field values**, consistent with a typical PDF viewer, when PDFium supports the form.

## Actual behavior

`out.png` **omits** the filled field text; the rest of the page rasterizes as usual.

## Screenshots

If applicable, attach:

Issue:

Image

Expexted:

Image

## Environment

- pdfplumber version: `0.11.9`
- OS: macOS

## Additional context

**Likely fix:** In `get_page_image`, after successfully opening `pypdfium2.PdfDocument`, call **`pdfium_doc.init_forms()`** before **`pdfium_doc.get_page(page_ix)`**, matching pypdfium2’s documented order.

**Related (not duplicate):** [#120](https://github.com/jsvine/pdfplumber/issues/120) is about form values not appearing in **CLI / extraction**; the README documents AcroForm via pdfminer. This report is specifically about **`to_image()` / `get_page_image`** rasterization.

Cyberfilo:
## Summary

\`page.to_image()\` (which goes through \`pdfplumber.display.get_page_image\`) currently opens a \`pypdfium2.PdfDocument\` and loads the page **without initializing PDFium's form environment**. Filled AcroForm field text is rendered via PDFium's form-rendering layer (\`FPDF_FFLDraw\`), which only runs when a form environment exists. Result: filled form-field text is **missing from the PIL bitmap**, even though every standard PDF viewer shows it.

This is purely a PDFium rasterization gap — pdfminer text extraction (\`page.chars\`, \`extract_text()\`) is unaffected.

## Reproduction

```python
import pdfplumber

with pdfplumber.open(\"filled_form.pdf\") as pdf:
    im = pdf.pages[0].to_image(resolution=150).original
    im.save(\"out.png\")  # filled field values absent
```

(Reproducer + sample PDF in #1367.)

## Fix

One-line addition in \`pdfplumber/display.py::get_page_image\`:

```python
pdfium_doc = pypdfium2.PdfDocument(src, password=password)
...
pdfium_doc.init_forms()  # <-- new
pdfium_page = pdfium_doc.get_page(page_ix)
```

Per [pypdfium2's documentation](https://pypdfium2.readthedocs.io/), \`init_forms()\` must be called **after** open and **before** loading pages. On a document without a form it is a no-op, so this is safe for the common non-form case.

## Test plan

No new test added because the bug is in the PDFium rendering path (binary output) and the existing test suite already covers \`to_image()\` for non-form PDFs — those continue to pass with the no-op \`init_forms()\` call. For form PDFs the existing tests cannot detect the regression visually; the rasterized bitmap differs only in a few hundred pixels per filled field.

Happy to add a pixel-comparison or perceptual-hash test against the sample PDF in #1367 if maintainers want — let me know.

## CHANGELOG

Added under \`## Unreleased\`.

Closes #1367.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file. The format 
 
 ## Unreleased
 
+- Fix `to_image()` / `get_page_image` omitting filled AcroForm widget values from the rasterized PIL bitmap. Calls `pypdfium2.PdfDocument.init_forms()` after open and before page load so PDFium's form-rendering layer (`FPDF_FFLDraw`) draws filled fields. Resolves [#1367](https://github.com/jsvine/pdfplumber/issues/1367).
 - Upgrade `pdfminer.six` from `20251230` to `20260107`. ([07a5ff6](https://github.com/jsvine/pdfplumber/commit/07a5ff6))
 
 ## 0.11.9 — 2026-01-05
--- a/pdfplumber/display.py
+++ b/pdfplumber/display.py
@@ -58,6 +58,12 @@ def get_page_image(
     except pypdfium2.PdfiumError as e:
         raise MalformedPDFException(e)
 
+    # Initialize the PDFium form environment before loading the page so
+    # filled AcroForm widgets are drawn by FPDF_FFLDraw at render time.
+    # init_forms() must run after open and before get_page(); calling it
+    # on a document without a form is a no-op.
+    pdfium_doc.init_forms()
+
     pdfium_page = pdfium_doc.get_page(page_ix)
 
     img: PIL.Image.Image = pdfium_page.render(


jsvine:
Hi @Cyberfilo, thank you for flagging this! I've merged #1376, which is a similar fix though also comes with a test.

PR #1375, opened 2026-06-05
Update Pillow dependency to address security concerns

tkjoetang:
This PR updates the minimum Pillow dependency from 9.1 to 12.2.0.

The change was motivated by security concerns raised during an internal dependency review. Older Pillow versions may be flagged by security scanners due to known vulnerabilities.

## Changes

* Updated Pillow dependency requirement
* Updated CHANGELOG.md
* Added myself to the contributors list

## Testing

Ran:

```bash
python -m pytest
```

Most tests passed locally. The Ghostscript-related tests in `tests/test_repair.py` were not validated because Ghostscript is not installed on my work machine.


jsvine:
Now handled in latest version. Thanks for flagging.

PR #1376, opened 2026-06-15
Initialize PDFium's form environment so filled AcroForm fields render in to_image()

### LINKED ISSUE #1367: `to_image()` / `get_page_image` omits filled AcroForm widgets because `PdfDocument.init_forms()` is never called

cycsmail:
Fixes #1367.

`get_page_image` opens the PDFium document but never calls `init_forms()`, so any filled AcroForm widget content is missing from the bitmap that `Page.to_image()` produces. Adding the call right after the document is opened (and before the page is loaded) fixes it; it's a no-op for PDFs that have no form, so non-form rendering is unchanged.

Added a regression test reusing the existing `federal-register-2020-17221.pdf` fixture: it renders page 0 and asserts that a crop over a filled form field has non-white pixels. Without the fix that region is blank (0 non-white pixels) and the test fails; with it the region has content (945) and it passes. The `tests/test_display.py` suite passes (17 tests) and black/isort/flake8 are clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 All notable changes to this project will be documented in this file. The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
+## [Unreleased]
+
+### Fixed
+- Initialize PDFium's form environment in `get_page_image` so that filled AcroForm field content is included when rendering pages via `Page.to_image()`. ([#1367](https://github.com/jsvine/pdfplumber/issues/1367))
+
 ## [0.11.10] — 2026-06-14
 
 ### Changed
--- a/README.md
+++ b/README.md
@@ -574,6 +574,7 @@ Many thanks to the following users who've contributed ideas, features, and fixes
 - [Brandon Roberts](https://github.com/brandonrobertz)
 - [@ennamarie19](https://github.com/ennamarie19)
 - [Anton Ilin](https://github.com/bronislav)
+- [Sebastian Cao](https://github.com/cycsmail)
 
 ## Contributing
 
--- a/pdfplumber/display.py
+++ b/pdfplumber/display.py
@@ -58,6 +58,12 @@ def get_page_image(
     except pypdfium2.PdfiumError as e:
         raise MalformedPDFException(e)
 
+    # Initialize the form environment so that filled AcroForm field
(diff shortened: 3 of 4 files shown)


jsvine:
Thanks @cycsmail — this seems like a helpful and concise fix, as well as a useful test. Much appreciated. Merging to `develop`.

Issue #1181, opened 2024-08-04
inconsistent coordinate systems when cropping


wodny:
## Describe the bug

It seems that `Page.crop()`:
- on one hand by default expects cropping coordinates compatible with the page's bbox (`strict=True`),
- on the other objects' coordinates are relative to the bbox.

For pages which have a bbox not starting at `(0, 0)` this causes `page.crop(page.bbox)` to return an empty set of objects. Adding `relative=True` does not help because it makes it two times worse. This is related to #245.

Note that externally provided PDF's may already be cropped. This is what `mutool` (MuPDF) does when using the `poster` function. It copies the page into multiple pages and then adjusts their `MediaBox`.

## Have you tried [repairing](https://github.com/jsvine/pdfplumber/blob/stable/docs/repairing.md) the PDF?

Repairing the PDF fixes the problem but:
- the original PDF does not seem to be broken; it contains unusual values but they seem to be legal according to PDF 32000-1:2008,
- this requires an external binary to be run which is far from perfect,
- there is an actual inconsistency in `crop()`.

BTW, `README.md` doesn't mention `repair`.

## Code to reproduce the problem

```python
import pdfplumber as pp

with pp.open("pages-cut-x.pdf") as f:
    p1, p2 = f.pages[:2]

    print("page 1 bbox", p1.bbox, "rects:", len(p1.rects))
    print("page 2 bbox", p2.bbox, "rects:", len(p2.rects))

    print()
    print("page 1 tables")
    for t in p1.find_tables():
        print(tuple(map(int, t.bbox)))

    print()
    print("page 2 tables")
    for t in p2.find_tables():
        print(tuple(map(int, t.bbox)))

    print()
    print("page 1 text")
    print(p1.extract_text_simple())
    print()
    print("page 1 text (cropped)")
    print(p1.crop(p1.bbox).extract_text_simple())
    
    print()

    print("page 2 text")
    print(p2.extract_text_simple())
    print()
    print("page 2 text (cropped)")
    print(p2.crop(p2.bbox).extract_text_simple() or "--- no text ---")
    print()
    # with strict=True:
    # ValueError: Bounding box (0, 0.0, 420.9449, 595.2756) is not fully within parent page bounding box (420.9449, 0.0, 841.8898, 595.2756)
    print("page 2 text (cropped non-strict by p1 bbox)")
    print(p2.crop(p1.bbox, strict=False).extract_text_simple())
    print()
[...]

PR #1381, opened 2026-07-09
Fix coordinate inversion when MediaBox origin is not (0, 0)

### LINKED ISSUE #1332: Problem with coordinate inversion when origin is not bottom left
## Describe the bug

It is possible to have pdf where the origin is not bottom left. For example, I have a pdf with centred origin. The raw mediabox is (-100, -200, 100, 200) and the resulting bbox in pdfplumber is (-100, 200, 100, 600).

## Expected behaviour

Following pdfplumber convention (origin at top left), bbox should be (0, 0, 200, 400), with all the content in the right system.

## Actual behaviour

If we have raw bounding box
- x0_r=-100
- y0_r=-200
- x1_r=100
- y1_r=200

We get that the page bbox is
- x0 = x0_r = -100
- y0 = -y0_r = 200
- x1 = x1_r = 100 
- y1 = (y1_r - y0_r) - y0_r = 600

## Screenshots

Image

## Environment

- pdfplumber version: 0.11.5
- Python version: 3.11.13
- OS: Mac

apoorvdarshan:
Fixes #1332.

### Root cause

`_invert_box` (in `pdfplumber/page.py`) flips PDF coordinates vertically so
the origin ends up in the top-left. It did so about the MediaBox **height**:

```python
def _invert_box(box_raw, mb_height):
    x0, y0, x1, y1 = box_raw
    return (x0, mb_height - y1, x1, mb_height - y0)
```

with `mb_height = mb_raw[3] - mb_raw[1]`. Because it subtracts from the
height rather than from the raw top edge (`mb_raw[3]`), the MediaBox's
raw y-origin (`mb_raw[1]`) is never subtracted. For a MediaBox whose
lower-left is `(0, 0)` this is fine (`mb_height == mb_raw[3]`), but when
the MediaBox does not begin at `y == 0` every `top`/`bottom` is shifted by
that origin, and `page.mediabox` / `page.bbox` start at a non-zero `y`.
(pdfminer.six already normalizes object coordinates to a `(0, 0)`-origin
page, so the flip must not re-introduce the origin.)

### Reproduction

A PDF with a centered-origin MediaBox `[-100 -200 100 200]` containing a
single rectangle drawn at raw coordinates `10 20 30 40 re` (raw span
`x: 10..40`, `y: 20..60`):

```python
import pdfplumber
with pdfplumber.open("centered.pdf") as pdf:
    p = pdf.pages[0]
    print("mediabox:", p.mediabox)
    r = p.rects[0]
    print("rect top/bottom:", r["top"], r["bottom"])
```

**Actual (before):**

```
mediabox: (-100, 200, 100, 600)
rect top/bottom: 340.0 380.0
```

The page's top edge sits at `top == 200` and `mediabox[3] == 600` even
though the page is only `400` units tall — the whole page is shifted down
by the origin (`200`).

**Corrected (after):**

```
mediabox: (-100, 0, 100, 400)
rect top/bottom: 140.0 180.0
```

`top` now ranges over `[0, height]`, as expected.

This also matches the report in #1332: raw MediaBox `(-100, -200, 100, 200)`
previously yielded `(-100, 200, 100, 600)`.

The regression is also visible on the existing `tests/pdfs/issue-1181.pdf`
fixture, whose first page has MediaBox `[0, 200, 420.9449, 585.2756]`:

- Before: `page.mediabox == (0, -200.0, 420.9449, 185.2756)`, and object
  `top` values were shifted `-200` into negative territory
  (`min(top) == -41.3`).
- After: `page.mediabox == (0, 0.0, 420.9449, 385.2756)`, and `min(top)`
  is `158.7` (all non-negative).

### Fix

Flip about the raw upper-`y` edge (`mb_raw[3]`) instead of the height.
This subtracts the origin correctly and is a no-op for the common case
(where the MediaBox lower-left is `(0, 0)`, so the top edge equals the
height). The change is confined to the two MediaBox/CropBox call sites in
`Page.__init__`; the x-offset behavior introduced for #1181 is preserved
(object x-coordinates and `page.mediabox[0]` still keep the raw
MediaBox `x0`), and downstream consumers (`process_object`,
`point2coord`) read the corrected origin from `self.mediabox`
automatically, so no other code needed to change.

### Tests

Added `tests/test_issues.py::test_issue_1332`, which uses the existing
`issue-1181.pdf` fixture (its first page has a non-zero MediaBox
y-origin) and asserts that the inverted MediaBox begins at `y == 0`, that
no object has a negative `top`, and that a sample character's coordinates
are corrected. The test fails on the current code
(`assert -200.0 == 0`) and passes with the fix. The existing suite
continues to pass (`python -m pytest`, excluding the `test_repair.py`
cases that require a local Ghostscript install, which is unrelated to this
change). `black`, `isort`, and `flake8` are clean on the changed files.

Disclosure: prepared with AI assistance; reviewed and verified locally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file. The format 
 ### Fixed
 - Initialize PDFium's form environment in `get_page_image` so that filled AcroForm field content is included when rendering pages via `Page.to_image()`. ([#1367](https://github.com/jsvine/pdfplumber/issues/1367))
 - Include the wrapped exception's class name in `PdfminerException`'s message when `pdfminer.six` raises an exception without one (e.g. `PDFPasswordIncorrect`), which previously surfaced as a blank error message.
+- Subtract the `MediaBox` origin when inverting coordinates, so that pages whose `MediaBox` does not begin at `y == 0` are no longer shifted vertically (previously producing negative `top` values). ([#1332](https://github.com/jsvine/pdfplumber/issues/1332))
 
 ## [0.11.10] — 2026-06-14
 
@@ -688,4 +689,3 @@ Whoops.
 
 ### Fixed
 - Fix find_gutters — should ignore `" "` chars
-
--- a/README.md
+++ b/README.md
@@ -577,6 +577,7 @@ Many thanks to the following users who've contributed ideas, features, and fixes
 - [Sebastian Cao](https://github.com/cycsmail)
 - [Kaspar Naraghi](https://github.com/kaninaba94)
 - [Siddharth Gaur](https://github.com/siddharthgaur1)
+- [Apoorv Darshan](https://github.com/apoorvdarshan)
 
 ## Contributing
 
--- a/pdfplumber/page.py
+++ b/pdfplumber/page.py
@@ -177,10 +177,15 @@ def _normalize_box(box_raw: T_bbox, rotation: T_num = 0) -> T_bbox:
 
 # PDFs coordinate spaces refer to an origin in the bottom-left of the
 # page; pdfplumber flips this vertically, so that the origin is in the
-# top-left.
-def _invert_box(box_raw: T_bbox, mb_height: T_num) -> T_bbox:
+# top-left. The flip is performed about the MediaBox's top edge
+# (`mb_top_edge`, i.e. the raw upper-`y` coordinate) rather than about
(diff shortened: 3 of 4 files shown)


apoorvdarshan:
Rebased onto current `develop`; preserved both current changelog entries and all contributor additions while resolving the overlaps. `tests/test_issues.py`: 25 passed.

Issue #1369, opened 2026-04-13
y-coordinates are "top to bottom" for words but "bottom to top" for line points.


SimonAndreys:
## Describe the bug

y-coordinates are "top to bottom" for words but "bottom to top" for line points.

## Minimal Working Example

Attached pdf shows a vertical line with "a" at the top and "b" at the bottom.

[lignes.pdf](https://github.com/user-attachments/files/26675498/lignes.pdf)

Attached code prints the coordinates of "a" and "b" and of x0, y0, x1, y1 of the only line in the pdf.

[plumbTest.py](https://github.com/user-attachments/files/26675781/plumbTest.py)

## Expected behavior

The coordinates of "a" should be within 10 points of the coordinates of x1,y1 (or x0,y0).

## Actual behavior

The y-coordinate of "a" do not match with y1 (nor y0). Instead it is close to page.height-y1, which correspond to a change from top-to-bottom direction to bottom-to-top.
'''
a coordinates:
 x0=71.04614, y0=127.08910000000003, x1=73.33113999999999, y1=132.08910000000003
line :
 x0=73.33114, y0=710.980737, x1=73.33114, y1=767.67365
line  (reverted y):
 x0=73.33114, y0=130.90906300000006, x1=73.33114, y1=74.21615000000008
'''

## Environment

- pdfplumber Version: 0.11.9
- Python 3.12.3
- OS : linux

###Remark
fixing this might be a breaking change for some applications...

Issue #845, opened 2023-03-22
hyperlinks have negative height


bentsi:
## Describe the bug
hyperlink height property has negative height value.

## Code to reproduce the problem
1) open pdf
2) see pdf_file.pages[61].hyperlinks

## PDF file
https://www.singtel.com/content/dam/singtel/about-us/sustainability/reports/Singtel-Group-Sustainability-Report-2022.pdf

## Expected behavior

height should be positive number

## Actual behavior

height has negative value

## Screenshots

![image](https://user-images.githubusercontent.com/5269241/226985552-e63f63fa-8238-4042-a949-150ea9531b13.png)

## Environment

- pdfplumber version: 0.7.5
- Python version: 3.10.5
- OS: Ubuntu 20.04.5 LTS (Focal Fossa)

## Additional context

in addition we can see that "top" and "bottom" attributes are swapped, that doesn't comply with pdfplumber's bounding box definitions as discussed in https://github.com/jsvine/pdfplumber/issues/198

jsvine:
Hi @bentsi, thanks for sharing this example. The `height`, `top`, and `bottom` attributes are all  calculated from the raw annotation's `Rect` (bounding box), specified by the PDF in a direct command. 

In this particular PDF (as observed by opening it in a text editor), that Rect command is `Rect[428.053 634.536 453.041 626.144]`, which corresponds to exactly what you see for `x0`, `y0`, `x1`, `y1` in your screenshot above, suggesting that `pdfplumber` is collecting the correct information.

Given that, there would seem to be two main options:

- Do nothing, on the principle that `pdfplumber` should focus on PDF objects' actual (i.e., as coded) attributes, rather than what we think the author intended.

- When `pdfplumber` sees an annotation that uses a bounding box that suggests a negative height, "fix" the bounding box (probably by flipping the vertical coordinates) so that it has a positive height. 

My inclination is toward the first option, because trying to fix PDF-creator's mistakes seems like opening a can of worms. But I'm open to suggestions otherwise.

soodoku:
This looks like it was fixed some time ago and the issue can probably be closed.

`Page.annots` now normalizes the raw `Rect` before inverting it, in `pdfplumber/page.py`:

```python
x0, top, x1, bottom = _invert_box(_normalize_box((*pt0, *pt1)), rh)
```

[...]

Issue #198, opened 2020-04-06
bbox inversion


ubmarco:
Hi,
this issue is about better understanding pdfplumber's bounding box (bbox) definition.
README.md states the following bbox definition `(x0, top, x1, bottom)`.
I'm confused why the bbox definition is not `(x0, y0, x1, y1)`.

I spent a lot of time understanding where my visual debugging issues came from and I'm not sure if all below considerations are correct, please tell me when I'm wrong.

This is the definition of each coordinate:
```
x0      Distance of left side of object from left side of page.
x1      Distance of right side of object from left side of page.
y0      Distance of bottom of object from bottom of page.
y1      Distance of top of object from bottom of page.
top     Distance of top of object from top of page.
bottom  Distance of bottom of the object from top of page.
```
As I'm having continuous confusion with this, I painted some ASCII art to enable a discussion:
```
PDF standard and pdfminer bbox               pdfplumber bbox                
*-page------------------------*              *-page------------------------*
|                             |              |                ^      ^     |
|                             |              |               top     |     |
|                             |              |                v      |     |
|        +-bbox----+          |              |        +-bbox----+    |     |
|<--x0-->|         |    ^     |              |<--x0-->|         |    |     |
|        |         |    |     |              |        |         |  bottom  |
|<-------|---x1--->|    |     |              |<-------|---x1--->|    |     |
|        +---------+    |     |              |        +---------+    v     |
|            ^          |     |              |                             |
|           y0         y1     |              |                             |
|            v          v     |              |                             |
*-----------------------------*              *-----------------------------*
```
The left is how pdfminer handles bbox for objects. The PDF standard defines the coordinate system origin of pages as the bottom left corner, so pdfminer complies with that.
The left is how pdfplumber handles bboxes.

I wanna have one definition of a bounding box which means I have to write conversion code around pdfplumber which I don't like. Any opinions on that?

jsvine:
Hi @ubmarco, and thank you for the question. It's a good one, with an explanation that's worth documenting here. These are some of the reasons I chose to use a coordinate system with the origin (`0, 0`) in the top-left corner, rather than the bottom-left corner:

- People read PDFs, generally speaking, from the top-left toward the bottom-right. Although it is contrary to the PDF standard, placing the origin in the top-left felt more intuitive to me and, I guessed, to newcomers.

[...]

PR #1387, opened 2026-07-27
Document the `y0`/`y1` vs `top`/`bottom`/`doctop` coordinate systems in README

### LINKED ISSUE #389: Explain `top`/`doctop`/`bottom` vs. `y0`/`y1` in README.md
See here:

- https://github.com/jsvine/pdfplumber/issues/198
- https://github.com/jsvine/pdfplumber/discussions/384

soodoku:
Closes #389.

Adds a `#### Coordinates` section to the README, directly above the per-object property tables, explaining the two vertical coordinate systems and how they relate.

The README currently defines `y0`/`y1`/`top`/`bottom`/`doctop` five separate times — once per object type — but never explains why there are two systems, which one bounding boxes use, or how to convert between them. That gap keeps generating issues: #198, #845, #1332, and as recently as this April, #1369, where a user reported as a bug that "y-coordinates are top-to-bottom for words but bottom-to-top for line points."

The section covers:

- which properties are measured from the page bottom (`y0`/`y1`) versus the page top (`top`/`bottom`) and the document top (`doctop`);
- that bounding boxes are `(x0, top, x1, bottom)`, not `(x0, y0, x1, y1)`, and that this is what `.crop(...)` / `.within_bbox(...)` expect;
- why the top-left origin was chosen, summarized from your answer in #198;
- the conversion `top == page.height - y1`, with the caveat that a non-zero `MediaBox` origin shifts it by `page.mediabox[1]`;
- that objects parsed from the PDF carry both varieties, while `.extract_words(...)` and `.search(...)` results carry only the top-down ones, and that `pts` are `(x, top)` tuples.

Every factual claim was checked against the library rather than against the existing prose:

- `top == page.height - y1` and `bottom == page.height - y0` hold exactly on a standard page, and fail by exactly `page.mediabox[1]` on `tests/pdfs/issue-1181.pdf` (`MediaBox` origin at `y = -200`), which is what the caveat records.
- The property inventory was confirmed by inspecting real objects: `char`/`line`/`rect`/`curve`/`image`/`annot`/`hyperlink` carry all seven; `.extract_words(...)` returns `top`/`bottom`/`doctop` and no `y0`/`y1`; `.search(...)` returns `top`/`bottom`.
- Using the reproduction attached to #1369, the documented rule accounts for the reported numbers exactly: read consistently in the top-down system, the two words land within 4 points of the line's endpoints, which is the agreement the reporter expected.

Docs-only, so no tests are added. `make lint` and `make tests` both pass locally (172 passed).

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,33 @@ Each instance of `pdfplumber.PDF` and `pdfplumber.Page` provides access to sever
 
 Each object is represented as a simple Python `dict`, with the following properties:
 
+#### Coordinates
+
+Horizontal coordinates are unambiguous: `x0` and `x1` are both measured from the left edge of the page. Vertical coordinates come in two varieties, and mixing them is a common source of confusion:
+
+| Property | Measured from | Direction |
+|----------|---------------|-----------|
+|`y0`, `y1`| Bottom of the page | Upward |
+|`top`, `bottom`| Top of the page | Downward |
+|`doctop`| Top of the *document* | Downward |
+
+The `y0`/`y1` pair follows the PDF specification's own coordinate system, and is retained for compatibility with `pdfminer.six`. Everything else in `pdfplumber` places the origin in the __top-left__ corner. In particular, bounding boxes are expressed as `(x0, top, x1, bottom)` — *not* `(x0, y0, x1, y1)` — which is what `.crop(...)`, `.within_bbox(...)`, and the visual-debugging methods expect.
+
+The top-left origin was chosen because ([#198](https://github.com/jsvine/pdfplumber/issues/198)) people generally read PDFs from the top-left toward the bottom-right; because measuring from the top is what makes `doctop` — the distance from the top of the *whole document* — able to distinguish objects at similar heights on different pages; and because most adjacent layout systems (SVG, Canvas, PIL/Pillow, which `pdfplumber` itself uses for visual debugging) also put the origin in the top-left.
+
+For a page whose `MediaBox` begins at the origin — the typical case — the two systems convert simply:
+
+```python
+y1 == page.height - top
+y0 == page.height - bottom
+```
+
+If the `MediaBox` does not begin at `(0, 0)`, both conversions are shifted by its vertical offset, `page.mediabox[1]`.
+
+Not every object carries both varieties. Objects parsed directly from the PDF — `char`, `line`, `rect`, `curve`, `image`, `annot`, and `hyperlink` — have all of the properties above. Objects that `pdfplumber` computes for you use only the top-down properties: the dicts returned by `.extract_words(...)` provide `top`, `bottom`, and `doctop` but no `y0`/`y1`, and those returned by `.search(...)` provide `top` and `bottom`. Relatedly, a `line`'s or `curve`'s `pts` are `(x, top)` tuples.
+
+Because object bounding boxes are normalized, `top` is never greater than `bottom`, and `height` (equal to `bottom - top`) is never negative.
+
 #### `char` properties
 
 | Property | Description |
@@ -576,6 +603,7 @@ Many thanks to the following users who've contributed ideas, features, and fixes
 - [Anton Ilin](https://github.com/bronislav)
 - [Sebastian Cao](https://github.com/cycsmail)
 - [Kaspar Naraghi](https://github.com/kaninaba94)
+- [Gaurav Sood](https://github.com/soodoku)
 
 ## Contributing
 
--- a/README.md
+++ b/README.md
@@ -156,20 +156,18 @@ Horizontal coordinates are unambiguous: `x0` and `x1` are both measured from the
 
 The `y0`/`y1` pair follows the PDF specification's own coordinate system, and is retained for compatibility with `pdfminer.six`. Everything else in `pdfplumber` places the origin in the __top-left__ corner. In particular, bounding boxes are expressed as `(x0, top, x1, bottom)` — *not* `(x0, y0, x1, y1)` — which is what `.crop(...)`, `.within_bbox(...)`, and the visual-debugging methods expect.
 
-The top-left origin was chosen because ([#198](https://github.com/jsvine/pdfplumber/issues/198)) people generally read PDFs from the top-left toward the bottom-right; because measuring from the top is what makes `doctop` — the distance from the top of the *whole document* — able to distinguish objects at similar heights on different pages; and because most adjacent layout systems (SVG, Canvas, PIL/Pillow, which `pdfplumber` itself uses for visual debugging) also put the origin in the top-left.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -144,6 +144,33 @@ Each instance of `pdfplumber.PDF` and `pdfplumber.Page` provides access to sever
...
+```python
+y1 == page.height - top
+y0 == page.height - bottom
+```
+
+If the `MediaBox` does not begin at `(0, 0)`, both conversions are shifted by its vertical offset, `page.mediabox[1]`.
+
+Not every object carries both varieties. Objects parsed directly from the PDF — `char`, `line`, `rect`, `curve`, `image`, `annot`, and `hyperlink` — have all of the properties above. Objects that `pdfplumber` computes for you use only the top-down properties: the dicts returned by `.extract_words(...)` provide `top`, `bottom`, and `doctop` but no `y0`/`y1`, and those returned by `.search(...)` provide `top` and `bottom`. Relatedly, a `line`'s or `curve`'s `pts` are `(x, top)` tuples.
+
+Because object bounding boxes are normalized, `top` is never greater than `bottom`, and `height` (equal to `bottom - top`) is never negative.
```
jsvine: I think we can probably cut this line. Or does it clarify something that is a common source of confusion? (Perhaps I'm just overlooking that.)

README.md
```diff
@@ -144,6 +144,33 @@ Each instance of `pdfplumber.PDF` and `pdfplumber.Page` provides access to sever
...
+|----------|---------------|-----------|
+|`y0`, `y1`| Bottom of the page | Upward |
+|`top`, `bottom`| Top of the page | Downward |
+|`doctop`| Top of the *document* | Downward |
+
+The `y0`/`y1` pair follows the PDF specification's own coordinate system, and is retained for compatibility with `pdfminer.six`. Everything else in `pdfplumber` places the origin in the __top-left__ corner. In particular, bounding boxes are expressed as `(x0, top, x1, bottom)` — *not* `(x0, y0, x1, y1)` — which is what `.crop(...)`, `.within_bbox(...)`, and the visual-debugging methods expect.
+
+The top-left origin was chosen because ([#198](https://github.com/jsvine/pdfplumber/issues/198)) people generally read PDFs from the top-left toward the bottom-right; because measuring from the top is what makes `doctop` — the distance from the top of the *whole document* — able to distinguish objects at similar heights on different pages; and because most adjacent layout systems (SVG, Canvas, PIL/Pillow, which `pdfplumber` itself uses for visual debugging) also put the origin in the top-left.
+
+For a page whose `MediaBox` begins at the origin — the typical case — the two systems convert simply:
```
jsvine, line 161: Since the `y`-flavored variables are the "original" ones, and `top`/`bottom` the conversions, I might suggest framing the code block below in those terms, instead of vice versa.

README.md
```diff
@@ -144,6 +144,33 @@ Each instance of `pdfplumber.PDF` and `pdfplumber.Page` provides access to sever
...
+
+| Property | Measured from | Direction |
+|----------|---------------|-----------|
+|`y0`, `y1`| Bottom of the page | Upward |
+|`top`, `bottom`| Top of the page | Downward |
+|`doctop`| Top of the *document* | Downward |
+
+The `y0`/`y1` pair follows the PDF specification's own coordinate system, and is retained for compatibility with `pdfminer.six`. Everything else in `pdfplumber` places the origin in the __top-left__ corner. In particular, bounding boxes are expressed as `(x0, top, x1, bottom)` — *not* `(x0, y0, x1, y1)` — which is what `.crop(...)`, `.within_bbox(...)`, and the visual-debugging methods expect.
+
+The top-left origin was chosen because ([#198](https://github.com/jsvine/pdfplumber/issues/198)) people generally read PDFs from the top-left toward the bottom-right; because measuring from the top is what makes `doctop` — the distance from the top of the *whole document* — able to distinguish objects at similar heights on different pages; and because most adjacent layout systems (SVG, Canvas, PIL/Pillow, which `pdfplumber` itself uses for visual debugging) also put the origin in the top-left.
```
jsvine: The `doctop` part of the explanation, although true, feels like a bit of a distraction, and slightly difficult to follow. Maybe just keep this paragraph to the first and third parts?

README.md
```diff
@@ -144,6 +144,33 @@ Each instance of `pdfplumber.PDF` and `pdfplumber.Page` provides access to sever
...
+
+Horizontal coordinates are unambiguous: `x0` and `x1` are both measured from the left edge of the page. Vertical coordinates come in two varieties, and mixing them is a common source of confusion:
+
+| Property | Measured from | Direction |
+|----------|---------------|-----------|
+|`y0`, `y1`| Bottom of the page | Upward |
+|`top`, `bottom`| Top of the page | Downward |
+|`doctop`| Top of the *document* | Downward |
+
+The `y0`/`y1` pair follows the PDF specification's own coordinate system, and is retained for compatibility with `pdfminer.six`. Everything else in `pdfplumber` places the origin in the __top-left__ corner. In particular, bounding boxes are expressed as `(x0, top, x1, bottom)` — *not* `(x0, y0, x1, y1)` — which is what `.crop(...)`, `.within_bbox(...)`, and the visual-debugging methods expect.
```
jsvine, line 157: I'd cut the `, and is retained for ...` part of the first sentence. I'd also cut the part inside the em-dashes.

README.md
```diff
@@ -144,6 +144,33 @@ Each instance of `pdfplumber.PDF` and `pdfplumber.Page` provides access to sever
 
 Each object is represented as a simple Python `dict`, with the following properties:
 
+#### Coordinates
+
+Horizontal coordinates are unambiguous: `x0` and `x1` are both measured from the left edge of the page. Vertical coordinates come in two varieties, and mixing them is a common source of confusion:
```
jsvine, line 149: I might reframe this slightly. It's less that the coordinates "come in two varieties", but rather that `pdfplumber` uses `top`/`bottom` but retains reference to the original `y0`/`y1` coords.

jsvine:
Thanks, @soodoku! I do think it's a good idea to clarify the coordinate system. I've made some specific suggestions in the line-comments above. More broadly: The location in the README feels a bit off, since it breaks up the discussion of object types. This might go better at the end of that section, right before Visual Debugging. What do you think?

soodoku:
Thanks for the helpful suggestions — I’ve addressed all three in `619c3c5`: removed the normalization sentence, reframed the conversions from `y0`/`y1` to `top`/`bottom`, and simplified the origin explanation by removing the `doctop` aside. `make lint` and all 172 tests pass locally. Happy to adjust further if needed.

jsvine:
Wow, quick work! I'm curious: Any chance this is an LLM responding / committing?

PR #1392, opened 2026-08-25
fix: keep layout superscripts aligned with baseline numbers

### LINKED ISSUE #1380: page.extract_text(layout=True) incorrectly reproducing text
Open attached .pdf, and look at the end of the page for this line:
Against Celsus (Origen), 327 , 328 *, 329 *

page.extract_text produces two lines instead:
```asterisks on the top line should appear after 328, 329; but are shifted left
                                         *    *                                   
                  Against Celsus (Origen), 327 , 328 , 329                         
```

Yes:
```
$ gs -o repaired.pdf -sDEVICE=pdfwrite page826.pdf
GPL Ghostscript 9.55.0 (2021-09-27)
Copyright (C) 2021 Artifex Software, Inc.  All rights reserved.
This software is supplied under the GNU AGPLv3 and comes with NO WARRANTY:
see the file COPYING for details.
   **** Warning: considering '0000000000 XXXXX n' as a free entry.
   **** Warning: considering '0000000000 XXXXX n' as a free entry.
Processing pages 1 through 1.
Page 1

   **** This file had errors that were repaired or ignored.
   **** The file was produced by: 
   **** >>>> Mac OS X 10.13.6 Quartz PDFContext <<<<
   **** Please notify the author of the software that produced this
   **** file that it does not conform to Adobe's published PDF
   **** specification.
```
Running the script on it, no change

*Please try running your code with `pdfplumber.open(..., repair=True)` before submitting a bug report.*
The script already does this

Trying to run thru pdf2txt from pdfminer.six gives this, which looks right, or at least better:
```
Against Celsus (Origen), 327 , 328 *, 329 * 
```
```
import pdb
import pdfplumber

if __name__ == '__main__':

    with pdfplumber.open("page826.pdf", repair=True) as pdf:

        page = pdf.pages[0]

        one_string = page.extract_text(layout=True)
        lines = one_string.split('\n')

        print("\nasterisks on the top line should appear after 328, 329; but are shifted left")
        print('\n'.join(lines[53:55]),'\n')

        print("still, coordinates look right:\n")
        words = page.extract_words(return_chars=True)

        print("left '*' starts at x0=297.0, and '328' below ends at 294.0 -- OK")
        print("line 53:",words[236]['text'],words[236]['x0'],words[236]['x1'])
        print("line 54:",words[243]['text'],words[243]['x0'],words[243]['x1'],'\n')

        print("rite '*' starts at x0=330.0, and '329' below ends at 327.0 -- OK")
        print("line 53:",words[237]['text'],words[237]['x0'],words[237]['x1'])
        print("line 54:",words[245]['text'],words[245]['x0'],words[245]['x1'],'\n')

```

## PDF file
page826.pdf, attached

## Expected behavior

One line, not two; or, at the least, correct x-positioning of the two asterisks

## Actual behavior

They are shifted left, towards lower x0 and x1, and (unless I am wrong) by different amounts

## Screenshots

[page826.pdf](https://github.com/user-attachments/files/29637923/page826.pdf)

## Environment

- pdfplumber version: 0.11.9
- Python version: 3.10.12
- Ubuntu 22.04

## Additional context

BetterAndBetterII:
## Summary

With `extract_text(layout=True)`, a superscript `*` sitting just above a baseline number (common footnote / citation markers) was clustered onto its own layout line. That line kept raw PDF x-columns while the baseline line drifted right, so the markers appeared shifted left of `328` / `329` (issue PDF: Against Celsus line).

Two small layout-only adjustments:
1. Widen vertical line clustering for `layout=True` with `max(y_tolerance, line_density / 2)` so near-baseline superscripts share a line with the text they annotate.
2. X-sort words on layout lines even when `presorted=True`, so extraction order cannot leave `*` before the number it belongs to. `use_text_flow` is unchanged.

## Test plan

- [x] `pytest tests/test_utils.py` (37 passed), including new synthetic superscript layout regression
- [x] Reproduced on the issue's `page826.pdf`: one line `Against Celsus (Origen), 327 , 328 * , 329 *`

Fixes #1380

### DIFF
--- a/pdfplumber/utils/text.py
+++ b/pdfplumber/utils/text.py
@@ -341,10 +341,18 @@ class WordMap:
             else sorted(self.tuples, key=lambda x: line_cluster_key(x[0]))
         )
 
+        # Keep superscripts (slightly smaller `top`) on the same layout
+        # line as the baseline so their x-position is not computed alone.
+        if layout:
+            line_density = y_density if line_dir in ("ttb", "btt") else x_density
+            line_cluster_tolerance = max(y_tolerance, line_density / 2)
+        else:
+            line_cluster_tolerance = y_tolerance
+
         tuples_by_line = cluster_objects(
             words_sorted_line_dir,
             lambda x: line_cluster_key(x[0]),
-            y_tolerance,
+            line_cluster_tolerance,
             preserve_order=presorted or use_text_flow,
         )
 
@@ -374,7 +382,7 @@ class WordMap:
 
             line_tuples_sorted = (
                 line_tuples
-                if presorted or use_text_flow
+                if use_text_flow or (presorted and not layout)
                 else sorted(line_tuples, key=lambda x: char_sort_key(x[0]))
             )
 
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -417,6 +417,76 @@ class Test(unittest.TestCase):
         with pytest.raises(ValueError):
             p.extract_text(layout=True, layout_height=300, layout_height_chars=50)
 
+    def test_extract_text_layout_superscript_asterisk_x_position(self):
+        """
+        Superscript '*' just above a baseline number (issue 1380) must keep
+        its x-position in layout=True output: after the number it annotates,
+        not shifted left on a previous line.
+        """
+
+        def make_char(text, x0, top, width=6.0, size=12.0):
+            return {
+                "text": text,
+                "x0": x0,
+                "x1": x0 + width,
+                "top": top,
+                "doctop": top,
+                "bottom": top + size,
+                "width": width,
+                "height": size,
+                "size": size,
+                "upright": True,
+            }
+
+        # Narrow glyphs to the left make line_len outrun x/x_density, which
+        # is what leaves isolated superscripts shifted left of "328"/"329".
+        baseline_top = 100.0
+        super_top = baseline_top - 3.4
+        chars = []
+        x = 10.0
+        for letter in "AgainstCelsus":
+            chars.append(make_char(letter, x, baseline_top, width=5.0))
+            x += 5.0
+        for i, digit in enumerate("328"):
+            chars.append(make_char(digit, 90.0 + i * 6.0, baseline_top))
+        chars.append(make_char("*", 111.0, super_top))
+        for i, digit in enumerate("329"):
+            chars.append(make_char(digit, 123.0 + i * 6.0, baseline_top))
+        chars.append(make_char("*", 144.0, super_top))
+
+        text = utils.extract_text(
+            chars,
+            layout=True,
+            layout_width=250,
+            layout_height=130,
+            layout_bbox=(0, 0, 250, 130),
+        )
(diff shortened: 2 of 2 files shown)


perigot827:
I reproduced a regression on head `d2fd6de017953297fb965622130f23d28b65d7fa`: the increased line tolerance merges two separate, nonoverlapping rows, even with explicit `y_tolerance=0`.

The test below uses 4-point characters on rows 5 points apart, leaving a 1-point vertical gap. It returns `['Alpha', 'Beta']` on base `4c64b92d5caccd71c645e98e0fabb0c4dba7ff45`, but `['Alpha Beta']` on this head. All four parameter cases pass on the base and fail on the head. The same result reproduces through `Page.extract_text` with an original PDF generated using ReportLab, so it is not limited to synthetic character input.

At default `y_density=13`, `max(y_tolerance, line_density / 2)` forces a 6.5-point tolerance. This also changes output when `use_text_flow=True`. With `layout=False` or `y_density=4`, both revisions preserve the rows in this example.

Could the superscript handling preserve the caller's ability to keep distinct rows separate? This regression test needs no additional dependency beyond pytest and the library:

```python
"""Regression coverage for pdfplumber PR 1392; no additional dependency."""

import pytest

from pdfplumber import utils


def make_char(text, x0, top):
    return {
        "text": text,
        "x0": x0,
        "x1": x0 + 2,
        "top": top,
        "doctop": top,
        "bottom": top + 4,
        "width": 2,
        "height": 4,
        "size": 4,
        "upright": True,
    }


@pytest.mark.parametrize("use_text_flow", [False, True])
@pytest.mark.parametrize("y_tolerance", [0, 3])
def test_layout_preserves_separate_nonoverlapping_rows(use_text_flow, y_tolerance):
    chars = [
        make_char(letter, 30 + 2 * index, top)
        for word, top in [("Alpha", 100), ("Beta", 105)]
        for index, letter in enumerate(word)
    ]
    text = utils.extract_text(
        chars,
        layout=True,
        layout_bbox=(0, 0, 200, 200),
        layout_width=200,
        layout_height=200,
        use_text_flow=use_text_flow,
        y_tolerance=y_tolerance,
    )
    assert [line.strip() for line in text.splitlines() if line.strip()] == [
        "Alpha",
        "Beta",
    ]
```

I also ran the existing `tests/test_utils.py`: 36 tests pass on the base and 37 on this head, including the new superscript case. I did not run the entire suite. This report was prepared with AI assistance and its reproduction was executed locally on both revisions.

Candidate amendment: [patch and reproduction notes](https://gist.github.com/perigot827/e615ed043b30ea1519fbb8f12c06be43). This experimental amendment applies to head `d2fd6de017953297fb965622130f23d28b65d7fa`. In local checks it preserves the issue PDF's superscript placement and passes 18 extended distinct-row regression cases. The candidate's full suite passed 190 tests with Ghostscript 10.08.0, and project lint passed. These additional results concern the candidate amendment; the base/head review results above are unchanged. The approach uses conservative geometry checks and is not a general semantic-line guarantee.

